PR #19, opened 2020-03-10
Connection protocol testdev

nodlesh:
Added Some Connection Protocol Tests.
Test Harness enhancements.
Dockerize the Test Harness.
New Quick Start Guide

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 venv/
 __pycache__/
 
+.vscode/launch.json
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -16,7 +16,7 @@ cd von-network
 ./manage start --logs
 ```
 
-# Determine the Ledger Browser IP to use in the URL for the Aca-py Agent in the next section
+### Determine the Ledger Browser IP to use in the URL for the Aca-py Agent in the next section
 in another shell...
 Make sure there is a von_von network
 ```
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,106 @@
+# Aries Agent Test Harness Quick Start Guide
+
+This guide quickly gets the Aca-py Agent BDD tests up and running. It utilizes separate docker containers for the Indy Ledger and Ledger Browser, the ACA-PY agents and backchannels, and the test harness itself. 
+
+### Prerequisites
+- Docker is installed and working on the host machine.
+- git is install and working with a github account.
+
+## Run the Ledger and Ledger Browser
+In one shell...
+```
+git clone https://github.com/hyperledger/indy-sdk.git
+git clone von network stuff
+cd von-network
+./manage build
+./manage start
+```
+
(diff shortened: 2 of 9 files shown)


swcurran:
Unfortunately, I'm seeing some problems and am not able get this running.  I think we are leaving you too much on your own about issuing we've solved. We need to get together to talk things over. Definitely not your fault.  We just need to see if we can help more.  These are all minor issues, but they need to addressed.

- preference would be that you forked this repo and then did PRs from branches in your fork.  That's the normal dev model we use.
- the 5 commits are not signed, which is not permitted.  See the DCO notice above. Obviously I forgot to make sure you knew that.  Likely that can be fixed - see the "All checks failed" link above and the "Details" on the DCO.
- I couldn't get the agents to run - the method of finding the genesis file didn't seem to work.  Not sure what the issue is, but the easiest solution is to likely just get this into a docker-compose file so that (a) we don't have to do that finding the docker gateway IP and (b) so that we can run everything in the same terminal session.  I think we can get some help on this as soon as we can get this running.

I'm sure things are working for you, so would really like to see a demo tomorrow/Wednesday of the tests and test cases!

swcurran:
Just read the rest of the Quick Start guides. Can't wait to get things running so we can try everything!

nodlesh:
Thanks. I did not know about the fork requirement. 
The DCO stuff is my fault. You had mentioned it in chat and I glossed over it not know what it was, and didn't follow up with the research. 
Can you tell me what error you are getting. I did a fresh pull of everything in a clean environment and all seemed to work fine. 
I think the docker compose is a great idea but I think it is a little pre-mature. We are not sure yet if the test harness will provision the agents are not, so it may look really different in the not too distant future. The way it is now is far easier than it was a few weeks ago. 
I'll see if I can fix the DCO stuff and create a fork and submit a new PR.

swcurran:
I guess that would help, eh?  The error message is: `Error retrieving ledger genesis transactions`

I also tried a number of variations on the LEDGER_URL other than what was suggested, including using localhost and Dockerhost, but all gave the same error.

FYI - why do I need to `git clone` the indy-sdk?  I didn't do that, inc case that matters.

swcurran:
Re: the docker-compose.  The quickstart should allow just go and run so I don't think it is early for that.  There needs to be docs about how the user would then use the environment - e.g. write their own backchannel, run their own agent, etc.

nodlesh:
I was under the assumption that in order to run the ledger you needed the indy-sdk. Does not the von-network use that? Ian had it in the original guide along with von-network. 
I have the DCO signing fixed. Now working on figuring out how to fork and move my branch commits to it.

swcurran:
No need to move these commits over - we can accept these now that they are signed.  Just in future, use a fork.  von-network embeds the indy-sdk, as do the ACA-Py agents.  No need to have it on your system.

nodlesh:
Ah yes you are right. indy-sdk was never copied into the container anyway. I'll remove that line.

swcurran (approved):

PR #20, opened 2020-03-24
Update the Quick Start Guide

WadeBarnes:
- Document how to use `von-network`s `manage` script to determine the DockerHost IP address.
- Add `--rm` to the containers so they are automatically removed when shutdown.
- Add the `DOCKERHOST` environment variable so the agents register their endpoints with it.
  - Agents were registering their endpoints using `host.docker.internal`, which does not work on Linux implementations of Docker, it's only available in Windows and Mac implementations.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -16,15 +16,14 @@ cd von-network
 ./manage start --logs
 ```
 
-### Determine the Ledger Browser IP to use in the URL for the Aca-py Agent in the next section
+### Determine the DockerHost IP address to use for the Aca-py Agent in the next section
 in another shell...
-Make sure there is a von_von network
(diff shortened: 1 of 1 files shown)


WadeBarnes:
@swcurran, This is just a first step to get it working for everyone by simply updating the docs.  This should resolve your issue running on Linux.

nodlesh (approved):

swcurran (approved):
Looks good.

swcurran:
BTW @WadeBarnes - it would nice to have the `./manage dockerhost` result be just the IP address.  There are some weird characters (probably newlines...) that were a pain to filter out. Nice to do it once in the script.  That's a general issue -- not specific to this.

PR #21, opened 2020-03-25
Automate the standard test runs

WadeBarnes:
- First pass at automation.  Get the existing tests to run in a more automated fashion.


swcurran (approved):
Looks good - nice work!  Still needs some clean up of instructions.  Also wondering why we are using Ian's ACA-Py repo vs. PyPi (or at least master from HL/ACA-Py).

PR #22, opened 2020-03-26
Update dependencies.

WadeBarnes:
- Point are official releases/repositories.
- Sneak in minor log formatting fix for manage script.


swcurran (approved):
I think looks good - might be an extra fail, but we need these updates regardless.

PR #40, opened 2020-05-07
Updated build handling to add selection of what to build; added support for PWD

swcurran:
Added support for the following:

- Valid agents are dynamically determined by looking for Dockerfiles in the backchannels folder.
- Play with Docker is supported with the assumption that either GENESIS_URL is passed in OR von-network is running in the same PWD session. In either case, when on PWD, a test is done to see if the URL can be resolved.
- Arguments can be added to the build command to select the (valid) agents to build.  On PWD, this allows bypassing the PWD build of VCX which takes a really long time... :-)

### DIFF
--- a/docker/manage
+++ b/docker/manage
@@ -1,5 +1,9 @@
 #!/bin/bash
 # set -x
+
+# Set of valid agents - add new agents here
+export VALID_AGENTS=$(ls ../aries-backchannels/Dockerfile.* | sed "s/^.*file.//")
+
 export MSYS_NO_PATHCONV=1
 export DOCKERHOST=${APPLICATION_URL-$(docker run --rm --net=host eclipse/che-ip)}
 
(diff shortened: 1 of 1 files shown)


swcurran:
FYI - PWD instructions:

- start a PWD session on [PWD](https://labs.play-with-docker.com/) or [PWV](http://play-with-von.vonx.io/) (or run it on your own system in an empty folder)
- Run these commands (assuming this PR has not been merged):

```
git clone https://github.com/bcgov/von-network
cd von-network/
./manage build
./manage start
cd ..
git clone https://github.com/swcurran/aries-agent-test-harness
cd aries-agent-test-harness/docker/
git checkout manage-build-updates
./manage build -a acapy
./manage run -d acapy
```

nodlesh (approved):

PR #41, opened 2020-05-08
Rename Alice to Acme to imply that Acme is an enterprise issuer/verify in most scenarios

swcurran:


swcurran:
Tested for build and run with Acapy and VCX.

nodlesh (approved):

PR #49, opened 2020-05-23
add initial gherkin test case for present proof protocol RFC 0037

AdamJLemmon:
Very much a work in progress to translate the RFC into a set of tests to cover the full protocol but wanted to get initial feedback here before I get too far down the track.  Thanks!

Signed-off-by: Adam J Lemmon <adamjlemmon@gmail.com>


swcurran:
Good stuff.  FYI - I've added some more documentation that is in a PR, but should be merged soon on writing tests.

Do you think we have to introduce Faber into this? Adding another participant into the test cases adds some complexity to the entire suite.  If it is good enough that Bob have a credential from Acme (vs. Faber), we would not need that complexity.   It's not that it can't happen, but that we want to be sure we need to do it before we add new participants.

On the other hand, it's giving me some ideas about how we might do this.

aries-test-harness/features/0037-present-proof.feature
```diff
@@ -0,0 +1,15 @@
...
+   @T001-API10-RFC0037 @P1 @AcceptanceTest @NeedsReview @wip
+   Scenario Outline: Present a proof to a verifier where the verifier has requested the proof
+      Given we have "3" agents
+         | name  | role     |
+         | Bob   | Holder   | 
+         | Faber | Issuer   |
+         | Acme  | Verifier |
+      And “Bob” has a credential issued from "Faber"
+      And “Bob and Acme” have an existing connection
+      And "Acme" sends a presentation request to "Bob"
```
nodlesh, line 12: Looks good. I believe this should be a "When"? We should always try to utilize the full gherkin syntax. Given does the preconditions and pre-steps before we get to the steps that are the actions that will result in the outcomes we want. The When are those actions. "Then" is used to show outcome of those When steps.

swcurran:
Closing

PR #53, opened 2020-05-29
Issue Credential Tests to Main

nodlesh:
Suite of Issue Credential Tests. Note, the @Indy tag is used for tests with Aca-py. The remaining tests will fail due to Aca-py not supporting starting from the Request Credential.


swcurran (approved):
Looks good.  Tested running suite - nice!

PR #54, opened 2020-06-02
Update dockerfiles and python code so docker image folders match repo for debugging

swcurran:
- made /aries-backchannel the PYTHONPATH in docker image
- in docker image, added folders for the base class (python/) and the backchannels (acapy/, vcx/, pico/)
- adjusted paths as necessary to access data and other resources (e.g. vcx agency)
- Run starts from /aries-backchannel
- Invoked program includes path to file -- e.g. "python acapy/acapy-backchannel.py"


swcurran:
Please give it a try @nodlesh to see if the debugging works.  @ianco, please take a look and see if the approach I took is OK and fits with the python model.

ianco (approved):
Looks good to me

PR #58, opened 2020-06-08
Happy Path test definitions for Present Proof Feature File

nodlesh:
@AdamJLemmon if you get a chance please give some feedback on these tests. I couldn't add you as a reviewer for some reason. Thanks!

### DIFF
--- a/aries-test-harness/features/0037-present-proof.feature
+++ b/aries-test-harness/features/0037-present-proof.feature
@@ -1,0 +1,94 @@
+Feature: Aries agent present proof functions RFC 0037
+
+   @T001-API10-RFC0037 @P1 @AcceptanceTest @wip @NeedsReview
+   Scenario Outline: Present Proof where the prover does not propose a presentation of the proof and is acknowledged
+      Given "2" agents
+         | name  | role     |
+         | Faber | verifier |
+         | Bob   | prover   |
+      And "Faber" and "Bob" have an existing connection
+      And "Bob" has an issued credential from <issuer>
+      When "Faber" sends a request for proof presentation to "Bob"
+      And "Bob" makes the presentation of the proof
+      And "Faber" acknowledges the proof
+      Then "Bob" has the proof acknowledged
+
+      Examples:
+         | issuer |
+         | Acme   |
+         | Faber  |
+
+   @T001.2-API10-RFC0037 @P1 @AcceptanceTest @wip @NeedsReview
+   Scenario Outline: Present Proof of specific types and proof is acknowledged
+      Given "2" agents
+         | name  | role     |
+         | Faber | verifier |
+         | Bob   | prover   |
+      And "Faber" and "Bob" have an existing connection
+      And "Bob" has an issued credential from <issuer>
+      When "Faber" sends a request for proof presentation to "Bob" for <proof>
+      And "Bob" makes the presentation of the proof
+      And "Faber" acknowledges the proof
+      Then "Bob" has the proof acknowledged
+
+      # Fyi, some of these proof examples may end up in the negative/exception test case. leaving them here for now.
+      # This proof list is free to grow as cases are discovered and raised.
+      Examples:
+         | issuer | proof       |
+         | Acme   | address2    |
+         | Faber  | zip         |
+         |        | city        |
+         |        | address1    |
+         |        | state       |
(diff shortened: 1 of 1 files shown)


aries-test-harness/features/0037-present-proof.feature
```diff
@@ -0,0 +1,94 @@
...
+      And "Faber" and "Bob" have an existing connection
+      And "Bob" has an issued credential from <issuer>
+      When "Faber" sends a request for proof presentation to "Bob"
+      And "Bob" makes the presentation of the proof
+      And "Faber" acknowledges the proof
+      Then "Bob" has the proof acknowledged
+
+      Examples:
+         | issuer |
+         | Acme   |
```
AdamJLemmon, line 18: Curious, will this require a 3rd agent to be spun up?

aries-test-harness/features/0037-present-proof.feature
```diff
@@ -0,0 +1,94 @@
...
+      Given "2" agents
+         | name  | role     |
+         | Faber | verifier |
+         | Bob   | prover   |
+      And "Faber" and "Bob" have an existing connection
+      And "Bob" has an issued credential from <issuer>
+      When "Faber" sends a request for proof presentation to "Bob"
+      And "Bob" makes the presentation of the proof
+      And "Faber" rejects the proof so sends a presentation rejection
+      Then "Bob" has the proof unacknowledged
```
AdamJLemmon, line 89: Worth a separate closing state to note this was `rejected` and is not still pending, for example?

AdamJLemmon (approved):

aries-test-harness/features/0037-present-proof.feature
nodlesh: @AdamJLemmon Yes, unfortunately. We are trying to limit the amount of agents we have to spin up, but I think two cases do need to be captured here. One where the verifier was also the issuer (Faber), and then one where the issuer is a 3rd party (Acme) to the present proof.

swcurran:
Not that it matters, but recommend that in the case you are having the issuer and verifier be the same party, I recommend that you use "Acme" in both roles (vs. Faber), if for no other reason then you can do that today, without waiting for updates to the `./manage` script to add Faber to the equation.

aries-test-harness/features/0037-present-proof.feature
nodlesh: That is a good question. There are a set of states in the 1.0 version of Present Proof, but nothing that indicates what that closing unacknowledged state actually is. I'm going to assume the state is null? We can verify once we get that step executing. Then raise it as a bug or go forward with it as it is. The 2.0 version has an "abandoned" state which I think is what it would be set to in this case. What ever that ending state is, i think it should be asserted in that last step. This way the same Gherkin can be used for 1.0 and 2.0 and only the state check would be different. I'm also hesitant on using/checking states in 1.0 version of the tests because the RFC makes no mention of what the states are. I had to reach into the 1.0 protocol code to find out.

nodlesh:
@swcurran Yes, will do that if I'm ready to execute before faber is spun up. To do that though, I'd have to add another test scenario instead of relying on the data table. Wanted it as Faber, because if this is running as a suite of tests and the connection and issue cred tests we executed before this suite, then Acme and Bob already have some connections and creds. As a verifier Faber would be more "fresh".

aries-test-harness/features/0037-present-proof.feature
```diff
@@ -0,0 +1,94 @@
...
+
+   @T001.2-API10-RFC0037 @P1 @AcceptanceTest @wip @NeedsReview
+   Scenario Outline: Present Proof of specific types and proof is acknowledged
+      Given "2" agents
+         | name  | role     |
+         | Faber | verifier |
+         | Bob   | prover   |
+      And "Faber" and "Bob" have an existing connection
+      And "Bob" has an issued credential from <issuer>
+      When "Faber" sends a request for proof presentation to "Bob" for <proof>
```
nodlesh, line 29: @swcurran Does this line make sense given what we talked about in the standup this morning? <proof> represents the types in the examples below, each one will probably have its own cred schema or maybe one schema with all?

swcurran (approved):
Verified the tests run.

PR #63, opened 2020-06-26
First Working Scenario Outline (2 test scenarios when executed) for Presentation

nodlesh:
Presentation Scenario Outline T001-API10-RFC0037 working.


swcurran (approved):
Looks good. Tested with and without Faber and saw expected results.

PR #65, opened 2020-07-04
aries framework dotnet backchannel setup

TimoGlastra:
This PR adds initial setup for the backchannel for .net

- add `dotnet.Dockerfile` that contains agent + backchannel
- add backchannel server with all endpoints throwing not implemented exceptions
- implement everything needed for the `When "{inviter}" generates a connection invitation` step to test the setup.


swcurran (approved):
Ran the build and run scripts - w00t!

I'm must say that I'm also pleased that the ./manage script that I helped with automagically picked up the new backchannel. Nice!

PR #69, opened 2020-07-07
Convert url mapping to include the topic + operation in the url …

ianco:
... rather than payload, for aca-py backend.  (Should also work for VCX and Pico but not tested.)

Also added support for url with & without the trailing slash for all url mappings

Note - kept the csv file (for now) that defines the allowable topic/operation combinations.  We can revisit this when we build in the swagger functionality.

### DIFF
--- a/aries-backchannels/python/agent_backchannel.py
+++ b/aries-backchannels/python/agent_backchannel.py
@@ -158,9 +158,16 @@ class AgentBackchannel:
 
         app = web.Application()
         app.add_routes([web.post("/agent/command/{topic}/", self._post_command_backchannel)])
+        app.add_routes([web.post("/agent/command/{topic}", self._post_command_backchannel)])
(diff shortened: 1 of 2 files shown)


swcurran (approved):
No impact.  Now to change the test calling?

Can we merge this now?

ianco:
The tests are also updated.  We can merge this now, it will break the .NET backchannel until they convert it over to the new urls

swcurran:
I only see updates in the backchannel for acapy/vcx.  What am I missing?

ianco:
https://github.com/bcgov/aries-agent-test-harness/pull/69/files#diff-ead3ad85667e11f46c1ce15f090dfc53

swcurran:
Doh...even on the second look I didn't notice the test-harness.  I was thinking it was the base class and specific.  :palm:

swcurran:
BTW - are we paying by the line of code?  That was a big change with very little code to show for it...

:-)

ianco:
$1 per line of code changed, $10k for knowing which lines to change

PR #70, opened 2020-07-13
 dotnet: get operation from url instead of payload

TimoGlastra:
This PR updates the dotnet backchannel to retrieve the operation from the URL instead of the body.

It also switches the JSON library used to the same Aries Framework .NET uses to be more consistent.


swcurran (approved):
Builds and runs. Errors where expected.

PR #72, opened 2020-07-14
dotnet: allow to debug backchannel inside docker container

TimoGlastra:
Allows to debug a dotnet backchannel while running inside the docker container. I added a new `Debugging.md` file which adds a (for now) very short guide on how to achieve this for VSCode.

Currently it has some caveats, due to bugs and things that still need to be implemented. This will be fixed in future PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 venv/
 __pycache__/
 
-.vscode/
 aries-agent-test-harness.code-workspace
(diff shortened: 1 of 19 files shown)


swcurran (approved):
Kudos on getting in-docker debugging supported and documented.  Nice!  No doubt that will speed things up.

As expected, same time for run, same pass/fail/skip stats.

PR #78, opened 2020-07-23
change cred_ex_id to credential_exchange_id

TimoGlastra:
I'm sorry for this small change, but it bothers me when I see it in the swagger ui.

When the backchannel client sends an operation for the issue-credential topic it includes the credential exchange id as `cred_ex_id`. However when it receives the credential exchange id in the body of a response it checks for `credential_exchange_id`.

This changes it to be all `credential_exchange_id` (as ACA-Py uses)


TimoGlastra:
Fixed in #79

PR #80, opened 2020-07-24
webhook calls for proof removed and into Aca-py backchannel

nodlesh:
Webhook calls for proof are now removed and into Aca-py backchannel. Some related code cleanup and streamlining was done as well.

fyi @TimoGlastra


swcurran (approved):
Ran through all tests

PR #81, opened 2020-07-28
persist logs after running tests

TimoGlastra:
This PR makes the CUTs a bit easier to debug. After running the logs of the docker containers will be persisted to the `.logs` folder. This folder is ignored, so it is only for local testing.

I also made a small change to the acapy backchannel to also print the logs of aca-py itself. Now only the backchannel logs were printed.


swcurran (approved):
LGTM and things still run...

PR #88, opened 2020-08-11
Proof Request requested_values changed to attributes

### LINKED ISSUE #87: use requested_attributes instead of requested_values
When the AATH client sends a `proof/send-request` operation to a backchannel it uses `requested_values` as attribute key for the requested attributes. However the Indy API's expect this to be `requested_attributes`.

I now see the ACA-Py backchannel transforms this parameter before sending it to ACA-Py. I know AATH is not Indy specific, but when another format than Indy will be used the whole `request_presentations~attach` will be different, so I think it will be easier to just use `requested_attributes`.

Maybe I'm missing something here?

Example:

```jsonc
{
   "connection_id":"<<connection_id>>",
   "presentation_proposal":{
      "@type":"did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/request-presentation",
      "comment":"This is a comment for the request for presentation.",
      "request_presentations~attach":{
         "@id":"libindy-request-presentation-0",
         "mime-type":"application/json",
         "data":{
            "requested_values":{ // --> change to `requested_attributes`
               "attr_1":{
                  "name":"attr_1",
                  "restrictions":[
                     {
                        "schema_name":"test_schema",
                        "schema_version":"1.0.0"
                     }
                  ]
               }
            }
         }
      }
   }
}
```

nodlesh:
All references to requested_values in proof requests are now called requested_attributes.

closes #87

### DIFF
--- a/CONFIGURE-CRED-TYPES.md
+++ b/CONFIGURE-CRED-TYPES.md
@@ -148,7 +148,7 @@ The following is an example of a simple proof request for one attribute with som
 ```
 {
    "presentation_proposal": {
-      "requested_values": {
+      "requested_attributes": {
          "address_attrs": {
             "name": "address",
             "restrictions": [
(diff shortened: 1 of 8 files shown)


swcurran (approved):
Looks good -- tests seem happy.

PR #91, opened 2020-08-17
dotnet: improved debugging in vscode

TimoGlastra:
This PR adds improved debugging for the .NET backchannel. It now allows to run multiple debuggers in VSCode at the same time. This makes it possible to debug and set breakpoints for the interaction between two (or more!) agents at both sides.

The config is a little bit less dynamic than I hoped, but VSCode doesn't allow it to be more dynamic. So nothing to do about that.

A small change to the `./manage` script is made to check if an agent is already running. This is needed for the debugging, which starts the containers used for debugging before we run the `./manage` script.


swcurran (approved):
Didn't try the debugging, but verified that the manage script changes are working.

PR #93, opened 2020-08-19
Non-breaking changes to support connectionless proof tests in the future

nodlesh:
These code changes do not make the connectionless tests work. However they don't break anything either, and will support those tests in the future.


swcurran (approved):
LGTM - tested

PR #105, opened 2020-10-01
javascript: add did, schema and cred-def topics

TimoGlastra:
Adds all operations related to the did, schema and credential-definition topics to the javascript backchannel


swcurran (approved):
Looks good - ran through the test suite.

PR #115, opened 2020-11-16
AATH Allure Reporting Integration

nodlesh:
Allure Reporting Framework integrated. 
New Test Backchannel API endpoint for backchannels to implement for GET agent version. 
Updated README and new TEST_REPORTING.md guide.


ianco:
I get this error running the ./manage script, and then it seems to hang:

./manage: line 333: /Users/icostanzo/Projects/aries-agent-test-harness/aries-test-harness/reports/environment.properties: No such file or directory

nodlesh:
Ah yes, that file has to exist first. Will check it in. 

The test harness is not going to output what it usually does when using Allure. It's going to just sit there until it finishes the run.

TEST_REPORTING.md
```diff
@@ -0,0 +1,73 @@
...
+./manage run -d acapy -r allure -t @AcceptanceTest -t ~@wip
+```
+ Running locally and not in a build pipeline/continuous integration system, you will need to install the allure framework in order to generate and display the html report. 
+```
+$ pip install allure-behave
+```
+To generate the html report and start an allure report server. 
+```
+cd aries-test-harness
+allure generate --clean ./reports
```
ianco, line 35: I get "allure:  command not found"

Is there an allure command line we also need to install?

ianco: https://docs.qameta.io/allure/#_installing_a_commandline

TEST_REPORTING.md
```diff
@@ -0,0 +1,73 @@
...
+```
+If keeping  a history and reporting trends over time is important to do locally, the history folder inside the allure-report folder that was generated by Allure, will have to be copied into the reports folder before the next execution of the `allure generate` command after another test run. 
+```
+cd aries-test-harness
+$ cp -r ./allure-report/history ./reports
+allure generate --clean ./reports
```
ianco, line 42: Probably want to add a .gitignore for the reports and allure-report directories so they don't inadvertently get checked into github

ianco: How are you viewing the generated report?  I tried opening the generated index.html in chrome but a lot of content was blocked because I was opening as "file://"

ianco (changes requested):

nodlesh:
@ianco Ah yes, there are a few things missing. 
for command line
`brew install allure`
to run the allure report server, something like this
`allure open -h 192.168.2.141 -p 54236`

I'll also update that gitignore.

ianco (approved):

PR #116, opened 2020-11-19
Added timestamp to T001-RFC011 presentation

nodlesh:
In an effort to fix a problem with a revocation test using a dotnet agent, timestamps in presentations were added there. It ultimately didn't fix the problem with dotnet. That agent has something wrong in this regard. Further investigation into the dotnet agent is required. 
Adding the timestamp enhancements to the test harness still should be committed at this time.


aries-test-harness/features/0011-0183-revocation.feature
```diff
@@ -99,7 +118,7 @@ Feature: Aries agent credential revocation and revocation notification RFC 0011
       Examples:
          | issuer | credential_data   | timeframe       | request_for_proof              | presentation                  |
          | Acme   | Data_DL_MinValues | -86400:+86400   | proof_request_DL_revoc_address | presentation_DL_revoc_address |
-         | Acme   | Data_DL_MinValues | -604800:now       | proof_request_DL_revoc_address | presentation_DL_revoc_address |
+         | Acme   | Data_DL_MinValues | -604800:now     | proof_request_DL_revoc_address | presentation_DL_revoc_address |
```
swcurran, line 121: Per the "best practices guidance" that we have defined, proof requests should NEVER use a range of values. They should only use a single point in time -- either "now()" or some time in the past.  So I think you should eliminate the example that has a range.

swcurran (dismissed):

aries-test-harness/features/0011-0183-revocation.feature
nodlesh: Are you sure we should remove those original timeframes? These work, and someone already may be using the intervals like this. Keeping them there may at least capture any desired or undesired change to revocation. I understand that if everyone follows the best practices in 441 that these will not be needed, but maybe we should wait until 441 is out there?

esune: I think a better action would be to ADD cases only using one time boundary: the new code in aca-py (that will be merged) should handle things no matter what, and adjust the `to` `from` values to what is allowed/desired.

swcurran:
IMHO they should be removed as the semantic meaning is essentially impossible to implement, and easily used incorrectly.  They may inadvertently work in some frameworks based on a specific implementation, but it's unlikely to be interoperable.

I would be happier if the frameworks rejected such a proof request with a Problem Report rather than trying to process them.

esune:
This seems fine, as long as we don't leave edge-cases uncovered. The proposed "best practices" (currently a WIP) states:
> The verifier MUST specify, as current INDY-HIPE 11 notes, the same integer EPOCH time for both ends of the interval. In effect, where the presentation request specifies a non-revocation interval, the verifier MUST request a non-revocation instant.

The tests should then be updated so that:
1. same `to`/`from` is accepted
2. different `to`/`from` is rejected
3. only `from` specified is rejected (edge case of 2)
4. only `to` specified is rejected (edge case of 2)

nodlesh:
Are you sure open ended revocation "instants" should be rejected? 

Shouldn't 441 be updated to include those rejections in 2, 3, & 4? If they are added, then those existing test scenarios should remain, but become exception test cases. That is why they are still there, wasn't sure what would happen in those cases, and if just a reclassification of the test scenario would suffice.

esune:
> Are you sure open ended revocation "instants" should be rejected?

I think this is a behaviour we should define and write in the rfc so that there is  reference. As it stands now, open-ended intervals would not be possible since both intervals need to have the same value.

> Shouldn't 441 be updated to include those rejections in 2, 3, & 4? If they are added, then those existing test scenarios should remain, but become exception test cases. That is why they are still there, wasn't sure what would happen in those cases, and if just a reclassification of the test scenario would suffice.

We may want a more explicit definition than what is there, but the way I read it right now a reclassification to exception test case may be enough.

nodlesh:
We need to be clear on what "reject" means. The presentation is set to verified=false, or there is an error returned on the send-presentation call?

esune:
Good point, I always forget that "small" detail. I appears that we may very well need to nail down the requirements before we can test them ;)

I am not sure how complicated it would be to properly expose the rejection to a third-party app by only using the problem report, rather than  (also) setting some additional details on the presentation-exchange record such as `verified: false` and `state: unprocessable`, or something among those lines.

swcurran:
Hey I was clear on what reject means :-) -- you send back a problem report.  I guess I could have included the message to say. "REJECTED: FROM and TO values differ"

nodlesh:
Ah yes, so you did. Sorry I missed that.

esune:
I was thinking more about how do we also notify the holder that the error occur, but am now realizing that it probably is out of scope right now, sorry about that.

TimoGlastra:
> Further investigation into the dotnet agent is required.

@nodlesh If you can give me some context maybe I can be of help?

nodlesh:
Hi @TimoGlastra 
There are two issues. Both issues are with T002-RFC0011.  This Revocation test is used with the dotnet agent as the Bob(prover) role. The test creates a credential, then revokes it, then issues a new credential as a replacement, then Bob presents this new credential. 
This is the test.
```
   @T002-RFC0011 @RFC0011 @P1 @critical @AcceptanceTest @Schema_DriversLicense_Revoc @Indy
   Scenario Outline: Credential revoked and replaced with a new updated credential, holder proves claims with the updated credential with timesstamp
      Given "2" agents
         | name  | role     |
         | Bob   | prover   |
         | Faber | verifier |
      And "Faber" and "Bob" have an existing connection
      And "Bob" has an issued credential from <issuer> with <credential_data>
      When <issuer> revokes the credential
      And <issuer> issues a new credential to "Bob" with <new_credential_data>
      And "Faber" sends a <request_for_proof> presentation to "Bob" with credential validity during <timeframe>
      And "Bob" makes the <presentation> of the proof
      And "Faber" acknowledges the proof
      Then "Bob" has the proof verified
```
The dotnet agent errors when bob sends the presentation. The data from the test to the backchannel looks like this.
```
{
   "requested_attributes":{
      "address_attrs":{
         "cred_id":"54e801a3-9e94-468e-bc3f-4ae40274a5ec",
         "timestamp":1605734425,
         "revealed":true
      }
   },
   "self_attested_attributes":{
      
   },
   "requested_predicates":{
      
   }
}
```
**The first issue** is, when the request is sent from faber to bob and the request, incorrectly, doesn't send a non-revocation interval, then we get a 500 error. The agent logs show this.
```
      SendProofPresentation {"requested_attributes":{"address_attrs":{"cred_id":"54e801a3-9e94-468e-bc3f-4ae40274a5ec","timestamp":1605734425,"revealed":true}},"self_attested_attributes":{},"requested_predicates":{}}
info: Hyperledger.Aries.Features.PresentProof.DefaultProofService[2000]
      ProofRecordId fc331147-476a-42e6-9f9e-9a59b4f9dbd8
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HM4C0UL916CH", Request id "0HM4C0UL916CH:00000001": An unhandled exception was thrown by the application.
Hyperledger.Indy.InvalidStructureException: A value being processed is not valid.
   at Hyperledger.Aries.Features.PresentProof.DefaultProofService.CreateProofAsync(IAgentContext agentContext, ProofRequest proofRequest, RequestedCredentials requestedCredentials)
   at Hyperledger.Aries.Features.PresentProof.DefaultProofService.CreatePresentationAsync(IAgentContext agentContext, String proofRecordId, RequestedCredentials requestedCredentials)
   at DotNet.Backchannel.Controllers.PresentProofController.SendProofPresentationAsync(OperationBody body) in /src/Controllers/PresentProofController.cs:line 149
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.TaskOfIActionResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeActionMethodAsync>g__Awaited|12_0(ControllerActionInvoker invoker, ValueTask`1 actionResultValueTask)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeNextActionFilterAsync>g__Awaited|10_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Rethrow(ActionExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeInnerFilterAsync>g__Awaited|13_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeFilterPipelineAsync>g__Awaited|19_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Awaited|17_0(ResourceInvoker invoker, Task task, IDisposable scope)
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.<Invoke>g__AwaitRequestTask|6_0(Endpoint endpoint, Task requestTask, ILogger logger)
   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
   at Hyperledger.Aries.AspNetCore.AgentMiddleware.Invoke(HttpContext aHttpContext, IAgentProvider aAgentProvider)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
```
If it is blowing up on not having a non-revocation interval for a revokable credential, then we should be getting an appropriate  error back instead of the 500. 

**Second issue** is, if the non-revocation interval (this is really a non-revocation instant based on the best practices in RFC 441) is given in the request, then the new credential is not verified. This same test configuration is verified if acapy is used as bob.

ianco:
@TimoGlastra you can see which test is failing here:

https://github.com/hyperledger/aries-agent-test-harness/runs/1434163228?check_suite_focus=true

TimoGlastra:
I was able to reproduce the error, but apparently debugging with VSCode and .NET is broken in macOS Big Sur at the moment so I'm not able to easily debug what's wrong here. 

Hoping that it'll be fixed soon, I'm gonna try again once that is resolved (https://github.com/microsoft/vscode-docker/issues/2498)

As a side note, I'll see if I can include the indy rust logs inside the backchannel as that'll definitely add more context to the vague `Hyperledger.Indy.InvalidStructureException: A value being processed is not valid.` error.

nodlesh:
@TimoGlastra Something to be aware of when you are able to debug. In regards to the second issue above, we are working through a similar issue with the same test case when using acapy. The problem started to manifest when the test was calibrated for the best practices in 441, making the non-revocation interval an instant. Acapy will not verify the credential in this scenario with the non-revocation instant as from=now and to=now. When the instant is changed back to an interval of a certain timeframe, ie. from=yesterday, to=now, the presentation will verify.  This behaviour, and failure, is the same in dotnet.  We are trying to narrow down what is going on here and I will report back when we know more.

swcurran:
@nodlesh -- please keep us informed as to what is going on with ACA-Py.  Unless the implementation in ACA-Py is extremely tricky, involving multiple calls to the ledger, I'd be pretty surprised if the from=yesterday to=now() worked in all but the most trivial case -- e.g. where there has only been one revocation made by the issuer. For example, I would be very surprised if the issuer made a revocation every hour for some number of hours and the proof request interval was for just a part of that time -- e.g. the last hour.

ianco:
@nodlesh  fyi github has screwed up your DCO's (the edits you made directly through the github UI aren't DCO signed off)

swcurran:
Uh-oh -- DCO problems @nodlesh .  Please fix asap.  That's not good to have those...  We just had a big mess in indy-sdk because of that.

nodlesh:
Do you guys know of a quick way to fix the DCO problems?

swcurran:
Yup -- ask @WadeBarnes to help.  That's the only thing I know.

ianco:
There are some ways to retro-fix DCO issues (e.g. https://stackoverflow.com/questions/13043357/git-sign-off-previous-commits, or google for other answers) but I find the most straightforward is often to just copy change over to a fresh branch

TimoGlastra:
aries-rfcs has some documentation on signing off previous commits. Always does the trick for me.

https://github.com/hyperledger/aries-rfcs/blob/master/contributing.md#how-to-sign-off-previous-commits

nodlesh:
DCO is still complaining about commits I just signed following the instructions Timo gave. I'm not sure where to go from here. :( 
@WadeBarnes any ideas?

WadeBarnes:
Working with @nodlesh directly on solutions.

nodlesh:
I think we've missed the time in which the GitHub actions kickoff to run tests but this looks ready now. So sorry for the DCO issues and the time to resolve.

swcurran:
Tests have run and they failed again.  Are you able to see the test results?

nodlesh:
Yes, it has to do with Andrew's change to aca-py startup. This PR will fix that.

ianco (approved):

aries-test-harness/features/0011-0183-revocation.feature
nodlesh: @swcurran  I'm resolving this conversation since changing the ranges is based on 441, which is not submitted yet. The state of the tests against such changes is in flux, so I don't know what to effectively change them to. My advice is to submit this and worry about the interval moving to an instant when 441 is done. Thoughts?

TimoGlastra:
@nodlesh I added #123 to add indy logging. This gives some more context to errors. I didn't resolve the error yet, but here is a bit more detailed error: 

```
21:35:12 Trace src/api/anoncreds.rs:2050 | prepare_result_1: >>> Err(IndyError { inner: 

RevocationState not found by id: RevocationRegistryId("HnEN3CKrjGUCPeLyB7iabE:4:HnEN3CKrjGUCPeLyB7iabE:3:CL:270:default:CL_ACCUM:32b837be-5df0-4ee8-aa14-79594ba95260")

Invalid structure }) 
```

Will continue to look into it

ianco (approved):

PR #122, opened 2020-12-07
Additional startup params for aca-py agent

ianco:

### DIFF
--- a/aries-backchannels/acapy/Dockerfile.acapy
+++ b/aries-backchannels/acapy/Dockerfile.acapy
@@ -29,4 +29,6 @@ RUN mkdir -p ./bin
 RUN ln -s /usr/local/bin/aca-py ./bin/aca-py
 ENV PYTHONPATH=/aries-backchannels
 
+RUN ./bin/aca-py --version > ./acapy-version.txt
+
 ENTRYPOINT ["python", "acapy/acapy_backchannel.py"]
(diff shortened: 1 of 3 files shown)


nodlesh:
@ianco these are new flags are they not? If so this will not be compatible with the current released version of acapy or older versions. We are pretty much saying AATH will now only work with master+. Should we be doing this?

ianco:
@nodlesh  good point ... I added them to be compatible with the latest updates in aca-py master, I'll re-test without having these flags in AATH

ianco:
... seems like these params are required now to start the agents on the latest `master` branch

Any suggestions on how to deal with these kind of version differences?

esune:
> Any suggestions on how to deal with these kind of version differences?

Tag and match releases between AATH and Aca-Py?

swcurran:
@nodlesh added collection of the ACA-Py (or, more general -- the Component Under Test) version information so that these could be handled with that.  I think that should be usable.  Versions 0.5.6 and less have the old way; Master (for now) and Version 0.6.0 and later use the new set.  Master is the tricky one, but it can be moved as the per version changes occur.

ianco:
Hmmm the backchannel gets the aca-py version number from the `/status` endpoint, however we need the version number before starting aca-py so we can conditionally add startup parameters ...

andrewwhitehead:
@ianco I'd suggest running `aca-py --version`

nodlesh:
Well yes, you are very right about that...

ianco:
I'll add this to the Dockerfile:

`RUN ./bin/aca-py --version > ./acapy-version.txt`

... and then update the backchannel to get the version # from the text file

nodlesh:
Or jam it into an environment variable from docker to get it from there in the backchannel.

ianco:
Added to dockerfile and check version on startup

Tested with latest aca-py release and with master branch

nodlesh (approved):

PR #129, opened 2020-12-18
RFC 441 test calibration

nodlesh:

Tests updated for best practices and logic changes in Revocation because of RFC 441.
Note: Test Scenario T004-RFC0011 will fail with at least acapy. At this time, the expectation of this scenario is to pass, but it does not.

### DIFF
--- a/aries-test-harness/features/0011-0183-revocation.feature
+++ b/aries-test-harness/features/0011-0183-revocation.feature
@@ -4,8 +4,8 @@ Feature: Aries agent credential revocation and revocation notification RFC 0011 
       Given "Acme" has a public did
       And "Acme" is ready to issue a credential
 
-   @T001-RFC0011 @RFC0011 @P1 @critical @AcceptanceTest @Schema_DriversLicense_Revoc @Indy
-   Scenario Outline: Credential revoked by Issuer and Holder attempts to prove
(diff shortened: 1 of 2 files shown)


ianco (changes requested):
I still get one test error:

Failing scenarios:
  features/0011-0183-revocation.feature:100  Credential revoked and replaced with a new updated credential, holder proves claims with the updated credential but presents the revoked credential -- @1.1

nodlesh:
@ianco Yes, that is the test mentioned above in the PR. T004-RFC0011.  I can put @wip back in so it doesn't run, but I'm going on what @swcurran has said a few times over the last little while; we may want failing tests to run so we can track when they are fixed. At this point (until someone tells me different) I know that test will fail, but I expect it to pass when the CUT is working correctly.

ianco (dismissed):

ianco:
OK if the test is correct and aca-py is incorrect, then I think it's ok to have the failing test.  (I.e. when aca-py is fixed then the test will pass)

I don't think the test should be marked wip unless it's the test itself that is still in progress.  It sounds like the test is correct.

nodlesh:
@ianco @swcurran I have set the test case in question to wip. Changing things to call /present-proof/records/{pres_ex_id}/credentials to get creds out of the provers wallet will affect more than this test. Impacts will extend to other tests and on other backchannel implementations. This will require more thought. I'll be adding a work item to the backlog to work on in the near future.

ianco (approved):

PR #133, opened 2020-12-23
fix for allure reporting old errors

nodlesh:

This submission fixes the allure reports. It was reporting errors from old runs, even if they passed. It was also showing multiple executions per one single execution. The fix was to set CHECK_RESULTS_EVERY_SECONDS: NONE in the docker-compose file, and to update the send-results script with a clean-results operation before the current results are sent. 

@esune @WadeBarnes The Allure Docker containers will need to be bounced with the updates in the docker-compose.yml in this commit. I do not know what other things @esune has done with the docker-compose so I wouldn't use the one here like it is unless it matches everywhere else in the file.


ianco (approved):

esune:
@nodlesh `CHECK_RESULTS_EVERY_SECONDS: NONE` was always/already set in our OpenShift-deployed version to support multi-project configurations so I expect this to be good to go as-is - just try pushing results the same way you do here and they will (should) show up.

I don't otherwise see significant changes from our deployed version.

PR #137, opened 2021-01-11
Updated first DID Exchange Test Step to make connection_id optional

nodlesh:

Updated the first `When` step in DID Exchange tests since afgo does not produce a connection_id at this point. Will update later steps for afgo when connection_id comes into play.

Also a bonus DID Exchange test is here that does a role swap. 
Implicit Invitation DID Exchange test included is wip.

FYI @murpheux

### DIFF
--- a/aries-test-harness/features/0023-did-exchange.feature
+++ b/aries-test-harness/features/0023-did-exchange.feature
@@ -15,3 +15,33 @@ Feature: Establishing Connections with DID Exchange RFC 0023
       And "Acme" sends complete to "Bob"
       Then "Acme" and "Bob" have a connection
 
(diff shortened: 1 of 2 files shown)


ianco (approved):
Looks good.

Note that we need to update the KGR file when we add new tests or the "acapy-full" workflow will fail (compares results from day-to-day).

I'll have to add docs for this somewhere ...

swcurran:
I'll let you know if/when I get the error messages to try out the update process.

ianco:
> I'll let you know if/when I get the error messages to try out the update process.

The "acapy-full" workflow should fail this afternoon.

I opened a new PR https://github.com/hyperledger/aries-agent-test-harness/pull/138 with some updated docs.

PR #139, opened 2021-01-14
Did exchange test dev acapy

nodlesh:
Final set of specific DID Exchance test scenario definitions including Exception tests. 
Most exception tests are defined but will not run (@wip). 
Two implicit invitation tests will continually fail until it is fixed in acapy. 
Explicit invitation tests will pass.
acapy KGR file is updated.


swcurran:
The tests ran -- loved the output.  Is this expected:

```
features/0011-0183-revocation.feature  ...F.F.........FFFFFFFFFF
features/0023-did-exchange.feature  ..FFFF.FFFFFFFFFFFFFF
features/0036-issue-credential.feature  .....FF
features/0037-present-proof.feature  ........F....FFF.
```

swcurran:
Just realized I ran without saying `-t ~wip`, so that's likely a big issue in what I did.  Trying again.

nodlesh:
Yes you will need to exclude the wip as always. 
Are you getting that output without the `-r allure` option? You shouldn't be.

nodlesh:
@murpheux the update you need to get the connection id in receive request is in this PR.  We can talk after it is merged.

swcurran (approved):

PR #149, opened 2021-01-28
Allure security public viewer

nodlesh:
This submission will work with the secure allure server with a public viewer of reports.

- The workflows will now push results to the current open Allure server, but also push to the new secure server. 
- A new script for sending results to the secure server exists. 
- All workflows have been updated to include this new secured send results to Allure.
- We will run both the public and the secured Allure servers in parallel for a few days to make sure everything is working on the secure side. When it is confirmed we can retire the public server. 
- **NOTE** To make the secured workflows work, someone needs to create two `secrets` on the Hyperledger/aries-agent-test-harness repo. I do not have the privileges for this. Adding secrets is in Settings. The names of the two secrets need to be `AllureAdminUser` and `AllureAdminPW`. Ask me for the values when creating.

Included in this submission is a new workflow for afgo. This should include afgo in the daily test runs. It will run DID Exchange Tests only. It will use afgo agents only. 

Also included in this submission are the fixes needed to solve the misreporting problem with Allure and our tests. Though Behave supports non-unique Scenario descriptors, Allure doesn't seem to. Since I added the DID Exchange Tests, I only copied existing tests and added a new tag. This had the side effect of Allure replacing a previous test with the same scenario descriptor with the latest one, resulting in missing test results. All scenario descriptors are now unique. 

- A new KGR file is also submitted here because of the scenario changes.


nodlesh:
FYI the new secure Allure Server URLs are 
UI: `https://allure-von.apps.silver.devops.gov.bc.ca`
API: `https://allure-von.apps.silver.devops.gov.bc.ca/api`

ianco (dismissed):

ianco (approved):

PR #152, opened 2021-02-04
afgo version update

murpheux:
in the light of recent error from Allure, updated this to allow test to pass


nodlesh (approved):
Looks good.

PR #158, opened 2021-02-15
javascript: add issue credential, present proof

TimoGlastra:
- Update Aries Framework JavaScript (will add javascript-master dockerfile soon)
- Add Issue Credential, Present Proof
- Add github actions runners for javascript-javascript acapy-javascript javascript-dotnet
    - I'm not sure if the allure project will automatically be created @nodlesh


swcurran:
@TimoGlastra -- looks good. It would be good to include a line about what tests that you expect to run with this, but I ran them all and same the ones working.

@nodlesh -- we have to look at the tags, I think.  It would certainly be good to have "revocation" as a tag, so we can say '-t ~@revocation` and not run any of those.

swcurran (approved):
Looks good ran through all the tests.

nodlesh:
> @nodlesh -- we have to look at the tags, I think. It would certainly be good to have "revocation" as a tag, so we can say '-t ~@revocation` and not run any of those.

We do have a @RFC0011 and a @RFC0183 tag. Those will remove the execution of revocation tests.

nodlesh:
Allure projects `javascript`, `acapy-b-javascript`, & `javascript-b-dotnet` created.

Issue #164, opened 2021-02-18
Suggestions for reorganizing the tags.


swcurran:
I generated a list of the tags in use for all of the scenarios and found that we had about a 100.  I think we can reduce that significantly by dropping a number of them and replacing them with the following subsets.  See the notes associated with each group.

The effort for this issue should include the following:

- reset all the tags on the scenarios
- adjust the runset tags accordingly
- document the range of tags and how they are used.

## Tags for scenario ID (sequential, no meaning):

A tag for each scenario, whose only purpose is to be able to run just a test or two from the command line.

@t001

# Tags for Scenario Metadata:

A tag to convey something about the scenario. `@wip` is probably the most important, but a couple of others may be valuable.

@NeedsReview
@wip
@NegativeTest

## Scenario Coverage Tags:

A tag for each RFC and AIP specifically tested in a scenario -- excluding "underlying" RFCs that are covered in essentially every test.  For example, RFC0011 (decorators) is used (I think) in a lot of scenarios and so _might_ be a candidate to remove (or not...).

@RFC0011
@RFC0023
@RFC0036
@RFC0037
@RFC0160
@RFC0183
@AIP10
@AIP20

## Tags for Temporary Issues:

A tag tied to a specific issue that should be addressed at some point. It needs to reference the repo in which the issue resides.

@issue:https://github.com/hyperledger/aries-cloudagent-python/issues/418

## Tags on test variants:

A tag for scenarios that implement a variant of a test for (usually) a given RFC.

@Indy
@revocation
@delete_cred_from_wallet
@MultiUseInvite
@ProofProposal
@SingleTryOnException
@SingleUseInvite

nodlesh:
Took a look at what is there. It seems that I would only be removing 4 tags dealing with priority. The rest are needed. The overall tag count isn't going to change all that much. Here are the updates that would be done.

Changes:
- Test IDs from @T001-AIP10-RFC0160 to @T001-RFC0160 (test ids need to be unique across all features so keeping the RFC number does that)

Removals:
- P1, P2, P3, P4 (These were the original priority categorization, these can be replaced with the Allure ones)
	- blocker, critical, normal, minor, trival must stay. They are used by Allure. 

Additions:
- @AIP10
- @AIP20

Tags to stay, but are not mentioned above:
- @AcceptanceTest - A functional test taken from directly from the language in the RFC
[...]

PR #165, opened 2021-02-19
test tag updates with some experiments

nodlesh:

Tags are streamlined based on #164.  For now, RFC numbers are left in the Test IDs. Lets look at it again later and judge the usefulness of keeping that rfc included. 

KGR file updated.

Also contained here is an experiment to see how Allure handles inherited tags and expanded feature descriptions. This experiment is in the RFC0023 test suite.

Tag documentation is not updated yet. That will come after the experimentation.


aries-test-harness/features/0011-0183-revocation.feature
```diff
@@ -22,7 +22,7 @@ Feature: RFC 0011 0183 Aries agent credential revocation and revocation notifica
          | issuer | credential_data   | request_for_proof              | presentation                  |
          | Acme   | Data_DL_MaxValues | proof_request_DL_revoc_address | presentation_DL_revoc_address |
 
-   @T001.1-RFC0011 @RFC0011 @P2 @normal @AcceptanceTest @Schema_DriversLicense_Revoc @Indy @DIDExchangeConnection
+   @T001.1-RFC0011 @RFC0011 @normal @AcceptanceTest @Schema_DriversLicense_Revoc @Indy @DIDExchangeConnection
```
swcurran: Questions -- as I understand this test, it's got an issue and a verify, so shouldn't it include RFC0036 and RFC0037?

Also, it has revocation, so shouldn't we have "@revocation", so that we can easily exclude all revocation tests?  I called that one out as one I would like to be able to run alone (e.g. "-t @revocation") or exclude (e.g. "-t @RFC0036 -t ~@revocation").

I'm not sure why RFC0011 is called out for that one -- attachments.  Attachments are used in many tests independent of what is used -- e.g. in any test using a VC.  Perhaps my question more specifically is why it would be associated as the "primary" RFC (eg. included with the T001 tag).

Also, wouldn't the @DIDExchangeConnection be just replaced with an "@RFC0023"?

aries-test-harness/features/0160-connection.feature
```diff
@@ -68,7 +68,7 @@ Feature: RFC 0160 Aries agent connection functions
                         # the agents under test implement acks.
 
 
-   @T003-AIP10-RFC0160 @SingleUseInvite @P2 @normal @ExceptionTest @WillFail @OutstandingBug..418..https://github.com/hyperledger/aries-cloudagent-python/issues/418 @allure.issue:https://github.com/hyperledger/aries-cloudagent-python/issues/418 @RFC0160
+   @T003-RFC0160 @SingleUseInvite @AIP10 @normal @ExceptionTest @WillFail @OutstandingBug..418..https://github.com/hyperledger/aries-cloudagent-python/issues/418 @allure.issue:https://github.com/hyperledger/aries-cloudagent-python/issues/418 @RFC0160
```
swcurran: Randomly picking this one --- shouldn't there be an @RFC0160 tag on this?  Or is this the inheritance experiment?

aries-test-harness/features/0160-connection.feature
```diff
@@ -85,7 +85,7 @@ Feature: RFC 0160 Aries agent connection functions
       When "Mallory" sends a connection request to "Acme" based on the connection invitation
       Then "Acme" sends a request_not_accepted error
 
-   @T004-AIP10-RFC0160 @SingleUseInvite @P2 @normal @ExceptionTest @WillFail @OutstandingBug..418..https://github.com/hyperledger/aries-cloudagent-python/issues/418 @allure.issue:https://github.com/hyperledger/aries-cloudagent-python/issues/418 @RFC0160
+   @T004-RFC0160 @SingleUseInvite @AIP10 @normal @ExceptionTest @WillFail @OutstandingBug..418..https://github.com/hyperledger/aries-cloudagent-python/issues/418 @allure.issue:https://github.com/hyperledger/aries-cloudagent-python/issues/418 @RFC0160
```
swcurran: One thing I noticed in seeing all the tags was that there were multiple formats for the "issue-related" tag -- in some cases "issue:<URL>" and then this format with "@OutstandingBug..".  I thought it would be good to use one predictable format.

swcurran (commented):
Added some questions.

aries-test-harness/features/0011-0183-revocation.feature
nodlesh: That is a good point on including RFC0036 and RFC0037. I've left the RFC tag represent the main goal of the test scenario, which is revocation. The issue and verify are used to accomplish the goal. If we do add them in, we should distinguish them from the main goal. I am also considering how to capture the RFCs that are indirectly tested because they are contained or used in the protocol under test. So there is the Main:RFC0011, Secondary:RCF0036, Secondary:RFC0037, & Indirect:RFCyadayada.   This isn't simplifying the tags though. 

I had thought RFC0011 was the Revocation RFC, but not looking at it I was wrong. Not sure how that happened. What is the Revocation RFC number? That would be used instead of the `revocation` tag.

on the DIDExchangeConnection tag, that is a code controller. The RFC number would mean  (at my current understanding of the tags) that RFC0023 is the goal of the test. It isn't, it is like 0036 and 0037 above.

aries-test-harness/features/0160-connection.feature
nodlesh: It is. It's at the end of the line.

aries-test-harness/features/0160-connection.feature
nodlesh: This is a good point. The WillFail and OutstandingBug are used in code and I output this info in the test results when using the normal Behave output. That allure.issue tag is new and puts a clickable url in the Allure reports.  I could check for the allure.issue tag now and remove that OutstandingBug tag.

Looking at this now, raises the question, this test may not fail when running on another agent. These WillFail and Issue tags will not correspond to a passing agent. Not sure how to handle this. Will have stew on that some more.

aries-test-harness/features/0160-connection.feature
swcurran: Got it -- thanks. Not crucial, but maybe have all the RFCs are the start -- maybe an order like this: 

<wip - if present> <ID> <AIP(s)> <RFC(s)> <other> <issue(s)>

aries-test-harness/features/0011-0183-revocation.feature
swcurran: I've always wondered by RFC0011 was referenced :-).  It confused me.

There is no revocation RFC.  There is a revocation notification RFC that will be in AIP 2.0 (that 0183), but 0011 is decorators.  As such, I would go with "revocation" vs. an RFC.

I would be against the "Main:" and "Secondary:" ideas because I want to use them to be able to run all the tests for a specific RFC.  We already have a mechanism to say what the main one is (in the ID tag), so I don't see why we need to specify further.

aries-test-harness/features/0160-connection.feature
swcurran: One thing we could do -- have a tag format of issue{:<agent>}:<number>, where the agent is optional.  That way, we could just use "acapy" or "afgo" to indicate that the issue is in those repos.  If not present, the issue is in AATH.  So examples:

@issue:acapy:418
@issue:418

The <agent> would use the same format as the .manage script.

aries-test-harness/features/0011-0183-revocation.feature
TimoGlastra: Indy HIPE 0011 is revocation, maybe that's where it comes from?

https://hyperledger-indy.readthedocs.io/projects/hipe/en/latest/text/0011-cred-revocation/README.html

nodlesh: Thanks Timo, I think that is exactly where I got the 0011 number. 

Should I prefix 0011 with something else, or just use RFC0183 as the reference? 

Should I add AIP20 for revocation? Right now it doesn't have an AIP assigned because it wasn't part of anything.

swcurran: Since the HIPE is Indy specific, and revocation is not, I would not recommend using RFC0011 (different RFC) or HIPE0011 (Indy-specific). For now, I think it best to just use "revocation" as the indicator.

For the AIP references, I suggest that AIP10 be used of the credential exchange protocols are RFC 0036/0037, and AIP20 if the protocols are RFC 0453/0454.

That allows, for example, an test agent to run tests for "AIP10" and not "revocation" if it doesn't support it.

aries-test-harness/features/0160-connection.feature
nodlesh: So this means the `@allure.issue:https...` tag must stay as is (can't put an agent in that tag or allure won't find it).
We will remove `@WillFail` & `@OutstandingBug..XXX..http...`. 
We will replace those two with `@issue:acapy:XXX`, or `@issue:XXX`.
The removal of `@WillFail` means that we won't list issues on tests unless they cause the test to fail. A test with an issue that doesn't fail has not happened yet, but I envisioned it when I created the `@WillFail` tag.

aries-test-harness/features/0160-connection.feature
nodlesh: Note that, if the tag inheritance works with allure, then the order, or where these tags are, will be taken care of by Behave/Allure. AIP and RFC tags will go up and tag at the feature level, and the tests will inherit those tags.

aries-test-harness/features/0160-connection.feature
nodlesh: Cut the issue tags down to just the allure.issue one. If that exists, then there is a message sent to tell users that this will fail because of this issue.

aries-test-harness/features/0160-connection.feature
nodlesh: Tags reordered.

aries-test-harness/features/0011-0183-revocation.feature
nodlesh: RFC0011 changed to @revocation. Test ids are changed to T001-HIPE0011

nodlesh:
This is ready to be merged if everyone is ok with it. There will be more tag edits if the tag experiment works, but I need to see it in the daily run first.

TimoGlastra:
LGTM. A lot cleaner with some of the tags removed / renamed

ianco (approved):
LGTM2

swcurran:
Last ~~whine~~ check --- I don't think HIPE0011 is relevant because it is just a description of how Indy revocation works vs. revocation in general.  I'd prefer to represent that as `@revocation` -- e.g. T000x-revocation.

nodlesh:
> Last ~whine~ check --- I don't think HIPE0011 is relevant because it is just a description of how Indy revocation works vs. revocation in general. I'd prefer to represent that as `@revocation` -- e.g. T000x-revocation.

Can we do that in the next PR? Just want to see it in Allure for an evening. Will then implement the experiment across the board, and update those revocation test names as you suggest. Though, HIPE0011 is where the tests were derived.

ianco:
> ... HIPE0011 is where the tests were derived.

If the tests were derived from the HIPE I think it's worth leaving the reference in, for the benefit of future generations ...

swcurran:
I disagree, but OK.  :-)

PR #168, opened 2021-02-25
initial send results for interop summary

nodlesh:

This PR opened to give exposure to the implementation to solve the pushing of allure results to the Aries Interop Summary Web Site. It is not ready to merge.

There are example report output to allow the implementation of the dynamic loading of the page based on this data. 

There is a template of the docs for acapy in which context will be given in user readable format on the website. 

There is the initial send reports script, that will be used to push summary reports to the web site. 

To run the script - `./send_report_interop_summary.sh --test_project acapy https://allure.vonx.io/api`

The initial design will be considered to do the following.
- The `./send_report_interop_summary.sh` will be integrated with the existing AATH GHAs that run tests and generate test results. It will run at the end of the GHAs. 
- The docs will have their own GHAs are are only fired when a file is updated in the repo. The GHA will then push this file to the web server somehow.


nodlesh:
@swcurran  Just want to make sure we are on the same page and clear. Can you review this, and make edits for clarity. Thanks.

**Story**
AS AN Aries Agent Framework Business Owner, I WANT to know the status of the framework's interoperability with other frameworks, SO THAT we can take steps (business and technical) toward greater interop in the protocols and roles the framework was designed for.

**Acceptance Criteria**

```
Scenario: Framework Business Owner views the interop page
Given a Aries Agent Business Owner
When they visit the Aries Agent Interoperability Web Page
Then they will see a page header that explains that the page is about Aries Agent Interoperability
And they will see in the header text and links about general AATH concepts
     e.g. the roles of each agent involved in the tests
And they will see sections represented with a section header for each runset configuration
And they will see in each runset section information that classifies the tests in the runset
    e.g. the frameworks AIP Versions, startup flags, and other text info 
And they will see in each runset section the current number of tests, passes, fails, etc. in some delightful form
And they might (or might not) see in each runset section a historical daily trend from previous runset runs
And they will see in each runset section a paragraph providing some context about the current state of the results
    e.g. why there are failures, etc.
And they will see in each runset section a link to the full Allure report for this runset configuration
And there will be a footer that might have text or links about AATH or anything else that might be helpful
```

The `index.md` page above is generated by a GHAction that is triggered by an event. Depending on the runset stats are generated, that event might be based on any merge into the repo, or it might be just be run daily.

The result of the execution is a markdown file that provides the information to satisfy the scenario above. The markdown file will be merged into the repo in the '/docs' folder of the repo. 

The data in the `runset` date above comes from either 3 or 4 places:

- the header and footer are files in the /docs folder and are just added directly
- the `.github/workflows/test-harness-<runset>.yml` file
   - See note below about the needed text data that may go into the yml file
- the `aries-test-harness/allure/The-KGR-<runset>.json` file OR the <runset> json data on https://allure.vonx.io
   - we'll start with the just the KGR data, since it's dead simple to get, but the Allure data could be used as well.

Ideally, I will be able to add the text details to be collected from the runset `workflows` files (above), but if that is not practical, they will come from a new file that is solely for that information per runset.

```
Scenario: There is a need to update something associated with the runset
Given a Test Developer
And they have updated some of the source data used in generating the `index.md` file
    - added, deleted or modified the KGR file for a runset because of a test CUD
    - added, deleted, or modified the runset workflows yml file
    - updated the text data about the runset (e.g. the overview or current status)
When the /docs index.md file is generated via GHAction
The GitHub Action will update the version of the /docs index.md file
And GitHub will publish the new version via gh-pages to aries-interop.info
```

```
Scenario: Runset is executed daily with no change to the results
Given the GHA is executing for a runset's daily run
When the runset output is the same as the known good results
The developers are not notified.
```

```
Scenario: Runset is executed with a change to the results
Given the GHA is executing for a runset's daily run
When runset output is different then known good results
The developers are notified that the runset results did not match the current results
And we expect the developers to do something about that.
```

What the developer does will vary (see list in the scenario above), but in all cases to resolve the issue will be one of the items mentioned above related to the files that the contribute to the `/docs` `index.md` file.

Developers should never update the /docs index.md file directly.

swcurran:
@nodlesh -- took a pass at updating the text above.

My plan for the adding the text parts is to add them as environment variables to the workflows.  That might not be long enough, and if not I would go with a separate file.

nodlesh:
Looks like you striped any references to updating the interop summary files in the last two scenarios, which is the only thing we care about in these scenario in the context of this PR. Unless you are saying those two do not do anything to support the interop page, but I think they do. Well the last one does, but the second to last (`Scenario: Runset is executed daily with no change to the results`) does not, and was there to state that nothing is done in relation to the interop page in that scenario.

swcurran:
My proposal is that we only update this interop page when the test summary data changes, which is when the tests and KGR data is updated.  The target for this page is not the developer maintaining the agents or monitoring the tests -- it's people interested in the overall interop.  As such, they are less interested in the day to day variations vs the overall current state.  And, if they are interested in the day to day, they just have to click the link to drilling into the recent run results.

Basically -- my claim is that I don't see the test results changing much day to day so updating this page daily is not helpful to our intended audience.  That's the purpose of the Allure results.

Whenever we (e.g. those maintaining the suite) see a change -- we should be:

- reviewing the difference.
- if appropriate, adding an issue with something (the test suite, one or more of the agents being tested)
   - An issue would be added to the test suite if the cause of the difference in results is not immediately obvious.
- if appropriate, updating the subjective description of the state of the runset ("On Feb. 26 we identified a problem with AF-Go and reported it vias issue https//.....". It is causing 5 tests to fail").  Or if the problem is fixed and tests are passing -- removing that text.
- updating the KGR file.  This is always done once the issue is identified.

nodlesh:
Given your summary generation over the weekend, and your use of Allure instead of the KGR files, are you still leaning on using the KGRs or Allure? 

I still have concerns over using the merge of a new KGR file as a trigger to generate a new summary for that runset. 
- A KGR is updated based on a reaction to a failure, not the inciting incident and therefore results that are currently active. 
- KGR files are merged for may reasons, and they are easy to get wrong. It could be that a KGR is merged, but the results later that day will fail and be different than what was merged, causing another merge the next day to fix it. In this case, when merged the KGRs are proposed, but not proven until the first run in the daily test execution. Updating the summary based on unproven KGRs seems incorrect. They also contain a lot of data. Just a change to a tag in an existing test causes a regen of the KGR. Actual results don't change in this case, therefore no reason to regen the summary. Imagine if all runsets use the KGR files, a minor test change that doesn't affect the pass/fail status, all KGRs have to be updated for every runset that runs that test. This will cause a lot of regen on the interop summary. 

My proposal, as it relates to KGRs is to have a GHA that is only manually triggered. Whomever merges the KGR monitors the run that evening to see if the KGR is proven to be correct. If so, they manual activate that GHA.  Another option, (I don't know GHA enough to know if this would work) We could have a GHA to activate on merge of a KGR, detect a change to the pass/fail status, and then "listen" for the test run GHA to send a proven status to it in the next run. At that point update the summary. 

On workflow merge, the GHA should only look at certain sections of the workflow for changes and only update the summary if the sections of concern changes.

<a side note on KGRs> Given my concerns above, implementing KGRs across all runsets is going to potentially cause a lot of maintenance.

swcurran:
It's not the merge of the KGR I'm proposing as the trigger of the run -- it's any merge.  Even then, I'm not particularly tied to that -- it could be a daily trigger.  I do think we should NOT change the file (merge it back) unless the file changes, but even that I'm not particularly hung up about.

The use of KGRs is different from the summary reporting, so let's keep the topics separate.  The purpose of the "notifications of GHA failure" is to get someone to do something.  Right now, the tasks are failing everyday (I'm getting notices, at least) and I don't see things being updated -- the errors are being ignored.  My assumption is that most runsets will have expected failures on a regular basis (e..g we expect that some tests will fail).  If we expect regular test run errors, then runs with expected errors should NOT trigger a notification. The only thing that should trigger a notification is when the runs are different than we expect.

PR #186, opened 2021-03-11
gitignore the environment.properties file

swcurran:
I keep having to restore the environment.properties file.  I'm assuming it should be ignored and so added it 
to the .gitignore file in this PR.  @nodlesh  -- please confirm that's ok to do.

I notice that there is a ".gitkeep" file in the /allure_results file, and that the file is mentioned in the
.gitignore file as well.  Is the intention that the folder needs to exist as empty in the repo but not 
to have any of the files in it saved to the repo?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ reports/
 allure-report/
 allure-results/
 New-KGR-File.json
+aries-test-harness/allure/allure-results/environment.properties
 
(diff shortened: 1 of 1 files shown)


nodlesh (approved):
Yes, local changes to the environment.properties can be ignored. 
There is no reason to have .gitkeep file anymore either.

PR #190, opened 2021-03-12
Issue cred v2 test dev

nodlesh:
Initial submission of working Issue Credential v2 tests for review. 

Tags are added at the scenario level to categorize indy, json-ld, json-ld-bbs, formats. Only Indy is working. To get json-ld test working one will need to update the data files along with possibly format_cred_proposal_by_aip_version. 

Before merge I still need to adjust runsets for this PR. Also need to regenerate the KGR file for acapy full.


aries-test-harness/agent_test_utils.py
```diff
@@ -52,3 +52,33 @@ def get_relative_timestamp_to_epoch(timestamp):
...
+def format_cred_proposal_by_aip_version(context, aip_version, cred_data, connection_id, filters=None):
+
+    if aip_version == "AIP20":
+
+        filters = amend_filters_with_runtime_data(context, filters)
+
+        credential_proposal = {
+            "connection_id": connection_id,
+            "credential_preview": {
+                "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/issue-credential/1.0/credential-preview",
```
TimoGlastra: Should be didcomm.org + 2.0 right?

```suggestion
                "@type": "https://didcomm.org/issue-credential/2.0/credential-preview",
```

TimoGlastra (dismissed):
Nice! 

I can't oversee at the moment if a lot of changes are required for the json ld credentials, but I think it should be fairly straightforward with the current approach

Two questions:
- I see you tagged the tests with json-ld and json-ld-bbs. How are the signature algorithm used going to be communicated? Is the json-ld scenario assuming ed25519 and is the json-ld-bbs (obviously) assuming bbs? ACA-Py is only supporting these two at the start, but AFGO supports more algorithms. Should each algorithm be a separate test case? Should the supported algorithms be queryable?
- Are we for now only testing a single attachment format at a time? E.g. it could be possible to send a proposal with both indy + json-ld attachment. After which an offer is sent with only indy. Or both.

aries-backchannels/acapy/acapy_backchannel.py
```diff
@@ -561,10 +599,71 @@ async def handle_did_exchange_POST(self, op, rec_id=None, data=None):
...
+
+    def check_aip_version(self, data=None):
+        # check if the data contains the aip_version. If so retrive it and set it localy, then remove it from data
+        if data and "aip_version" in data:
+            self.aip_version = data["aip_version"]
+            data.pop("aip_version")
+            if len(data) == 0:
+                data = None
+        else:
+            self.aip_version = "AIP10"
```
swcurran: Per discussion last week, AIP shouldn't really be a setting on this, but rather the version of the credential exchange protocols being used (issue-credential/v2.0 and present-proof/v2.0).  The RFC being used could also be used as major versions also mean new RFCs.

nodlesh: Yes, good catch. That method is left over from the previous approach I was implementing. It is not being used anymore and should be removed. I'll do that.

swcurran:
For Timo's questions, my thoughts are as follows:

> - I see you tagged the tests with json-ld and json-ld-bbs. How are the signature algorithm used going to be communicated? Is the json-ld scenario assuming ed25519 and is the json-ld-bbs (obviously) assuming bbs? ACA-Py is only supporting these two at the start, but AFGO supports more algorithms. Should each algorithm be a separate test case? Should the supported algorithms be queryable?

I think we have json-ld-bbs as separate set of tests, because of the ability to do selective disclosure. I think for JSON-LD, we'll also have to have different tests with different algorithms, a test agent can choose to execute the tests (and hence algorithms) it supports.  Not sure how best to convey from the test harness to the backchannel the algorithm.

> - Are we for now only testing a single attachment format at a time? E.g. it could be possible to send a proposal with both indy + json-ld attachment. After which an offer is sent with only indy. Or both.

Let's start with just one attachment at a time and then we'll later implement a test that has multiple at a time -- once test components implement that functionality.

aries-test-harness/agent_test_utils.py
nodlesh: Yes, true about @type. Thanks. 

Right now the tests are structured to only do one format per scenario. I think that is a decent start. Using the tags, a framework could construct a set of tests they support. After we have the single format for a single scenario working well. It won't be hard to create scenarios that allow multiple format tags and will allow to combine in one scenario. 

For the proposal and offer scenario you mention, we should be able write the steps you are talking about like this. 
```
@T00X-RFC0453 @critical @AcceptanceTest @DIDExchangeConnection @CredFormat_Indy @CredFormat_JSON-LD @Schema_DriversLicense_v2
  Scenario Outline: Issue a Indy credential with the Holder beginning with a proposal of a indy and json-ld credential
    Given "2" agents
      | name | role   |
      | Acme | issuer |
      | Bob  | holder |
    And "Acme" and "Bob" have an existing connection
    When "Bob" proposes a "indy + json-ld" credential to "Acme" with <credential_data>
    And "Acme" offers an "indy" credential with <different_credential_data>
    And "Bob" requests the "indy" credential
    And "Acme" issues the "indy" credential
    And "Bob" acknowledges the "indy" credential issue
    Then "Bob" has the "indy" credential issued
```

Not sure if that answers all your questions or not? I don't know what the signature algorithm means, is that foreign to the RFC? If needed to differentiate, we will have to add it to the Scenario tag(s), or integrate it into the gherkin step definition.

nodlesh:
This should be ready for merge. Adjusted based on reviews. Acapy-full runset has a new KGR file with new tests. That runset also removes WIP tests from the KGR and workflow.  This should not affect other runset workflows since they already exclude DID Exchange based connections, which these tests use.

swcurran (approved):
Ran through the tests -- looks good to me.﻿

PR #201, opened 2021-03-24
add global context initialization

TimoGlastra:
The main change of this PR is adding a global context initialization that sets up the context.

Two reasons:
- As a starter it was a bit unclear which context is used, for what. The `environment.py` now initializes all context with a description of the structure of that context variable
- A lot of checks can be removed if we have a base context initialized. (e.g. the following:

```
if not hasattr(context, 'connection_id_dict'):
    context.connection_id_dict = {}

# Check for responder key existing in dict
if requester not in context.connection_id_dict:
    context.connection_id_dict[requester] = {}

context.connection_id_dict[requester] = {context.responder_name: resp_json["connection_id"]}
```

is now reduced to:

```
context.connection_id_dict[requester][context.responder_name] = resp_json["connection_id"]
```

Besides that it does some cleanup and fixes some typos. 

This PR doesn't change any functionality, or any of the tests. It's just a restructure and cleanup.

Just some thoughts: Looking at all context variables in a row, I think maybe we can consolidate some of them? Instead of each item having it's own context variable, we can maybe separate them into two groups. Where one is global data that can be used for multiple scenario's (e.g. issued schema's, credential definitions, public dids, etc...) and a scenario specific context that will be thrown away after each scenario


nodlesh:
I didn't look deep into this PR, but off hand, Behave has the concept of a Feature Context(think global) and the Scenario Context. The Scenario Context is only in scope per scenario, so no need to clear these things. The Feature Context is only in scope for each feature. That is how I understand it.

TimoGlastra:
Ah thanks for clarifying that. So if I understand correctly, no global context is used at all?

I've updated the PR and moved all context setup to the `before_scenario` method. If we now want to make something global in the future (e.g. reusing connections between scenarios that don't test connection setup) we can move it to the `before_all` method. Is that okay?

swcurran:
What is the status of this?  @TimoGlastra -- I wasn't sure about merging and @nodlesh was away.  Could you resolve the merges and we can try to get this merged.  Is this going to break anything in the runsets?  :-)

TimoGlastra:
I've updated the branch. 

I'd like @nodlesh to review and see what he thinks.

nodlesh:
The content of this PR seems good. I like the changes. They don't break anything, I've executed multiple runsets with no ill effects. I did think there was one problem, but I that is due to another PR @TimoGlastra submitted three days ago. PR #209 That PR has caused T001-RFC0453 to start failing, and the two new JSON-LD tests fail as well, but that may be expected. Timo can you look what was changed to make T001-RFC0453 fail? It is a change in the `@when('"{holder}" acknowledges the "{cred_format}" credential issue')` step for Issue Credential v2.

TimoGlastra:
Woops my bad! 

I changed the return type for credential id to:

```python
credential_id = resp_json[cred_format]["credential_id"]
```

instead of:

```python
resp_json["cred_ex_record"]["cred_id_stored"]
```

This was mainly because you can have multiple credential ids for a single credential exchange. So I moved them to the credential format specific object. But this is only present in our cwu-bbs branch, not ACA-Py core.

I'll see to make a temporary fix in the current ACA-Py backchannel until our code is merged

TimoGlastra:
Seems this is a bit outdated now. I'm happy to update the PR if it will be merged.

Are there any blockers on this PR apart from it being outdated?

ianco:
I think this PR is good, sorry for the delays in getting it reviewed and merged :-(

If you can get it up-to-date we'll get it approved and merged.

swcurran:
Close this?  Worth getting done?  Last chance :-)

TimoGlastra:
I'll be working on AATH starting next week again. I'll have a look then and see if this is worth updating or not.

TimoGlastra:
Quick!! Before it is outdated again! :-)

nodlesh (approved):
LGTM

PR #208, opened 2021-03-31
ACA-Py Issue Credential V2 - JSON-LD credentials

TimoGlastra:
This adds support for testing Issue Credential V2 JSON-LD Credentials.

- One new API endpoint added `issue-credential-v2/prepare-json-ld`
- `@ProofType_XXX` and `DidMethod_xxx` tags added to indicate did method and signature suite to use for test

The tests are working but still require some files to be in this repo to get it working. Once the ursa bbs signatures wrapper we wrote is released on PyPi this PR can be updated for merging.


Moopli:
@TimoGlastra what's the ETA on this?

We're looking to test Issue Credential V2 interop soon, and this PR is a prerequisite :)

TimoGlastra:
We need to get some things fixed in ACA-Py before we can merge this. 

Maybe it'd be better to split the test harness code from the ACA-Py backchannel. This way you can already start with the AFGO implementation.

Would that help you?

Moopli:
Absolutely :+1:

TimoGlastra:
Allright. The PR should be ready by tomorrow

aries-backchannels/acapy/requirements-bbs.txt
```diff
@@ -0,0 +1 @@
+aries-cloudagent[indy, bbs]@git+https://github.com/animo/aries-cloudagent-python@cwu-bbs-main
```
ianco: I think we should just add this to acapy-main (requirements-main.txt, Dockerfile.acapy-main) rather than adding a new "bbs" version of the build

ianco: (for me acapy-main was breaking without the ursa bbs library installed)

ianco (changes requested):

aries-backchannels/acapy/requirements-bbs.txt
TimoGlastra: That's the idea! Will update the PR

ianco (approved):
Confirm that all the AATH tests pass with this PR

PR #209, opened 2021-04-06
Issue Credential V2 - JSON-LD credentials

TimoGlastra:
This contains just the test harness part of the issue credential v2 json-ld code. The ACA-Py backchannel stuff will be added in #208

### DIFF
--- a/aries-backchannels/data/backchannel_operations.csv
+++ b/aries-backchannels/data/backchannel_operations.csv
@@ -70,6 +70,7 @@ rfc,command,response,topic,method,operation,id,data,description,id_details,data_
 0036 Issue Credential,X,,issue-credential,POST,store,Y,Y,Send credential acknowledgement with ID ,cred_exchange_id,"offer = {
     ""credential_id"": ""issue-credential exchange id string""
 }",cred_exchange_id
+0453 Issue Credential v2,X,,issue-credential-v2,POST,prepare-json-ld,,Y,Prepare for issuing a JSON-LD credential,,"data = { ""did_method"": ""key"", ""key_type"": ""ed25519"" }",did
 0453 Issue Credential,X,,issue-credential-v2,POST,send-proposal,,Y,Send a credential proposal,,"proposal = {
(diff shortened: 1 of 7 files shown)


TimoGlastra:
@Moopli this should be enough to issue json-ld credentials using the test harness. Let me know if I broke something, was a bit hard to test without the backchannel.

swcurran (approved):
I can't test this without building the BBS+ Branch (which I don't know how to do yet...), so will just approve and merge this PR.

swcurran:
Just as a matter of interest -- @TimoGlastra -- do you know why this command on this PR branch (and on master after merging):

 `./manage run -d afgo-master -t T001.1-RFC0453 -t T001.2-RFC0453` 

would result in no tests being run and skipping all of the features, scenarios and steps?

Seems weird...

TimoGlastra:
> I can't test this without building the BBS+ Branch (which I don't know how to do yet...), so will just approve and merge this PR.

That's the nice thing of docker. Just like the rest :)

```sh
./manage build -a acapy-bbs
LEDGER_URL_CONFIG=http://test.bcovrin.vonx.io TAILS_SERVER_URL_CONFIG=https://tails.vonx.io ./manage run -d acapy-bbs -t @CredFormat_JSON-LD
```

> would result in no tests being run and skipping all of the features, scenarios and steps?

Seems expected. AFAIK the tags are AND. So by specifying two tags from different tests no test will match. I tried on the commit before this PR and it also doesn't run any tests.

If we somehow extend the `-t` param to accept spaces (should be doable), we could write statements like:

```
-t "@T001.1-RFC0453 or @T001.2-RFC0453"
```

--- 

Just checked, this works:

```
./manage run -d afgo-master -t T001.1-RFC0453
```

swcurran:
Thanks, Timo.

Spaces in the -t parameters is quite difficult -- it requires some fun bash handling.  Because the parameters are passed through multiple layers, it's hard to get the quoting right.  I even put a note about that in the documentation.  

Hmmm...maybe we can come up with a way to AND/OR tags and actually pass to behave just a single "-t" parameter.

We do need to add a note about the ANDing of the parameters.

PR #210, opened 2021-04-07
Revocation for acapy dotnet runset

nodlesh:
New Revocation tests that mix non-revocable credentials with non-revocation intervals as well as mixing revocable credentials with no non-revocation intervals. 

T013-HIPE0011 will fail in acapy until sklump does his PR for the fix. 

T013-HIPE0011 will fail in the acapy -b dotnet runset. The test is a non-revocable credential with a proof request/presentation with a non-revocation interval. The error is an indy error `Revocation Registry Id not found`. Should this be logged in aries-framework-dotnet?

KGR file is updated for acapy-full runset.

@ianco The tag MobileTest is added to T013-HIPE0011.


nodlesh:
Framework dotnet issue mentioned above is logged. https://github.com/hyperledger/aries-framework-dotnet/issues/184

swcurran (approved):
LGTM

PR #215, opened 2021-04-14
Fixes for the new sent top level in cred defs and schema responses

nodlesh:

This submission fixes test issues introduced by aries-cloudagent-python PR #1080. Most tests will now pass, however I believe there is an issue/bug with DID Exchange send-request when a Public DID is used to establish the connection. Will need to talk to @sklump to determine. 
In the meantime this PR should be merged to fix the majority of tests.


swcurran (approved):
LGTM

Issue #174, opened 2021-03-02
Issue Credential V2 Test Development (Happy Path)


nodlesh:
Approach: https://app.zenhub.com/workspaces/von---verifiable-organization-network-5adf53987ccbaa70597dbec0/issues/hyperledger/aries-agent-test-harness/171

RFC: https://github.com/hyperledger/aries-rfcs/tree/master/features/0453-issue-credential-v2

PR #222, opened 2021-04-20
Present proof v2 test dev

nodlesh:
2 Working Indy formated Present Proof V2 test cases. 
Updated KGR for Acapy-full runset. 

Further work may need to be down to support JSON_LD etc. @TimoGlastra can assess and make recommendations when he works with it. 

The power of the Example tags to simplify the feature test definitions are shown here. One could change credentials types and credential formats in an example instead of a new test. There is a new tag (actually a renamed tag) `@CredProposalStart`, which was once the `@Indy` tag.  It denotes that when a credential is created as part of a test, that issue credential starts with a proposal. Without that tag, it would start with a request. 

This PR is related to #174


nodlesh:
I'm unsure if reclassifying the steps for V2 from 
`And "Bob" has an issued credential from <issuer> with <credential_data>`
to 
`And "Bob" has an issued credential with formats from <issuer> with <credential_data>`
is as clear as it should be. Any recommendations in contrast too `with formats` is welcome.

ianco (approved):
Looks good.  One question - is the backchannel_operations.csv file checked into github in two places?

nodlesh:
Yes, it is. One is generated, and one is moved for pushing into the agent container... I think. We can streamline that.

PR #224, opened 2021-04-29
KGR fixes for the new json-ld tests

ianco:


nodlesh (approved):
👍

TimoGlastra:
What are these KGR files? I'm still not sure what it does :-)

swcurran:
"Known Good Results".  The idea is that on the daily runs, we include tests we know will fail (since we do that) and we raise a GHA exception (triggering an email) whenever the run produces different results than in the KGR files.  That way, we are not notified when any given test fails -- only when the results change. 

Right now, we aren't sending any GHA messages after test runs, regardless of what happens in the test runs.

TimoGlastra:
Ah that's smart. I like it

PR #227, opened 2021-05-01
Fix summary results script and create GHA

ianco:


swcurran (approved):
Nice!  Is 45 after a good time to generate?  I would suggest about 6 hours offset from the running of the rest.

PR #236, opened 2021-05-14
fix: small error fixes for acapy-afgo interop

Moopli:
- fixed some environment variables that had dashes
- add leading / to a path in afgo backchannel, so urls are constructed
  correctly
- return empty string instead of dict in a stub that should return a string
- don't skip processing empty strings in utils/output_reader


nodlesh:
Hi @Moopli Would it be too much trouble to pull the afgo_backchannel changes in this PR? I'm working on a major refactor on the afgo_backchannel and those lines no longer exist anyway. I'll have a PR with an updated afgo_backchannel on Monday if that is ok.

ianco (approved):

PR #244, opened 2021-05-26
Refactor test harness to integrate auxiliary services, and add a local uniresolver service

Moopli:
Von-network and indy-tails-server are now 'services' that are automatically
started by the ./manage script when it runs tests. Additional services can
be added just like these, by creating a folder for the service inside
`services`, and creating a wrapper script in the style of the existing ones.

Adds a universal-resolver service with a sov driver, for agents to be able to
resolve published did:sov DIDs from the local VON network.

Adds agent-type-specific environment configuration, using an <agent-name>.env
file in the agent's folder.

Configures afgo-interop to use the local uniresolver for did:sov.

Refactors ./manage script to add new commands:
- service: subcommands to `start`/`stop` a service, view `logs`, delete&`clean`
- start: start services and agents
- test: execute tests on running agents
- stop: stop agents (but leaves services running, for faster subsequent tests)

`./manage run [agents] [other args]` is the same as:

    ./manage start [agents]
    ./manage run [other args]
    ./manage stop

### DIFF
--- a/.github/workflows/test-harness-acapy.yml
+++ b/.github/workflows/test-harness-acapy.yml
@@ -27,6 +27,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-afgo.yml
+++ b/.github/workflows/test-harness-acapy-afgo.yml
@@ -28,6 +28,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-afj.yml
+++ b/.github/workflows/test-harness-acapy-afj.yml
@@ -28,6 +28,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-aip10.yml
+++ b/.github/workflows/test-harness-acapy-aip10.yml
@@ -26,6 +26,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-aip20.yml
+++ b/.github/workflows/test-harness-acapy-aip20.yml
@@ -26,6 +26,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-dotnet-javascript.yml
+++ b/.github/workflows/test-harness-acapy-dotnet-javascript.yml
@@ -31,6 +31,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-dotnet.yml
+++ b/.github/workflows/test-harness-acapy-dotnet.yml
@@ -30,6 +30,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-full.yml
+++ b/.github/workflows/test-harness-acapy-full.yml
@@ -30,6 +30,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-afgo-acapy.yml
+++ b/.github/workflows/test-harness-afgo-acapy.yml
@@ -28,6 +28,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-afgo.yml
+++ b/.github/workflows/test-harness-afgo.yml
@@ -27,6 +27,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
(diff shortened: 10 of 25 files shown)


TimoGlastra (dismissed):
Nice, LGTM

How does this work if you use external services? I often run with the public BCGov tails server and ledger by setting environment variables. Does that still work?

Moopli:
@TimoGlastra looks like I should amend my PR so it checks if those environment variables are set, to determine whether it should run the local servers, I'll make that change.

nodlesh:
I like it. A few comments. Is it possible to make this backwards compatible with the previous way of executing? 
Wondering if we can use Von Network and tails server if already running. This will be needed to keep the daily test runs in  GitHub Actions as they are implemented. Otherwise when this is merged they will all break. 

This seems to work for afgo, but trying acapy the agents fail to start up with the old ./manage run command. I didn't try the other agents so not sure if they still work.

ianco:
Agree with @nodlesh .  I tried running `./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip` and everything started up fine but the services didn't shut down after the tests.  `./manage stop` stops the agents but doesn't stop the services.  I'd expect either (a) I'm already running the services, so no need to start or stop them, or (b) I'd like the `./manage` scrpt to both start and stop the services when I do `./manage run ...`

Moopli:
@nodlesh That's odd, acapy runs fine on my end, whether with `./manage start` or `./manage run`. And it is indeed possible to use the von network and tails server that are already running, by setting `LEDGER_URL_CONFIG` and `TAILS_SERVER_URL_CONFIG` to point to the local instances. I pushed another commit to add those environment variables to all the jobs in github actions, does that solve the issue?

Moopli:
@ianco 
> the services didn't shut down after the tests. ./manage stop stops the agents but doesn't stop the services. I'd expect either (a) I'm already running the services, so no need to start or stop them, or (b) I'd like the ./manage scrpt to both start and stop the services when I do ./manage run ...

I see your point - I considered it a courtesy since the services would need to continue running if you want to do multiple tests, but I agree that it doesn't scan as obvious clean behaviour. It would be easy to make `./manage run` stop any services it starts, however, so I could do that.

Moopli:
@ianco does my latest commit address your issue?

ianco:
Taking a look now ...  

based on your latest updates, if I'm already running von-network and the tails server, I have to provide the env vars `LEDGER_URL_CONFIG` and `TAILS_SERVER_URL_CONFIG` correct?

.github/workflows/test-harness-acapy.yml
```diff
@@ -27,6 +27,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
```
ianco, line 32: I don't think "localhost" will work here, it needs to be the internal IP address of the docker network (in the `manage` script we set `$DOCKERHOST`, which comes from:

`export DOCKERHOST=${APPLICATION_URL-$(docker run --rm --net=host eclipse/che-ip)}`

ianco (changes requested):

ianco:
> @ianco does my latest commit address your issue?

Starting and stopping services is working when I `./manage run ...`, thanks for the fix!

However I can't get the test to run when passing the `*_CONFIG` parameters.  When I run:

`LEDGER_URL_CONFIG="http://localhost:9000" TAILS_SERVER_URL_CONFIG="http://localhost:6543" ./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip`

... then the `./manage` script detects that the ledger and tails server are running, however the agents aren't able to connect to "localhost", they need the docker IP address.  (However if I pass in the docker IP address instead of localhost, the manage script can't detect that the ledger and tails server are running.)

Moopli:
@ianco
Just pushed a fix: created some script-wide variables - `_INTERNAL` for the URL to use within docker containers, and `_HOST` for the URL to use from the host. If you set `_URL_CONFIG` variables to localhost, then the `_INTERNAL` variables will replace `localhost` with the dockerhost IP. If you set `_URL_CONFIG` variables to external URLs, then the `_HOST` and `_INTERNAL` variables will be the same. ~~(unless the external URL has `localhost` in it...)~~

Edit: just pushed a fix in case an external url was silly enough to have localhost inside it (while not being a localhost url). The `sed` replacement now replaces `://localhost:` with `://${DOCKERHOST}:` so it _only_ applies to a bona-fide localhost url.

.github/workflows/test-harness-acapy.yml
Moopli: my latest commit fixes this as well - `localhost` in these URLs will be replaced with `$DOCKERHOST` for the URLs passed to the agents.

nodlesh:
I'm not using afgo-interop. I'm trying to be sure this works with straight acapy (and other agents). So I get the following... 
```
LEDGER_URL_CONFIG="http://localhost:9000" TAILS_SERVER_URL_CONFIG="http://localhost:6543" ./manage run -d acapy-main -t @AcceptanceTest -t @AIP20 -t ~@wip
Agents to be used:
  Acme - acapy-main
  Bob - acapy-main
  Faber - acapy-main
  Mallory - acapy-main

waiting for ledger to start
waiting for tails server to start
Starting Acme Agent using acapy-main-agent-backchannel ...
Starting Bob Agent using acapy-main-agent-backchannel ...
Starting Faber Agent using acapy-main-agent-backchannel ...
Starting Mallory Agent using acapy-main-agent-backchannel ...

waiting for Acme agent to start
waiting for Bob agent to start..............................\e[31m
The agent failed to start within 30 seconds.
\e[0m

waiting for Faber agent to start...............................\e[31m
The agent failed to start within 30 seconds.
\e[0m

waiting for Mallory agent to start..............................\e[31m
The agent failed to start within 30 seconds.
\e[0m
```
As you can see some agents start and others do not. It seems random which ones will start on any given run. 
I can run any of these agents using my normal command to run them when I'm developing tests with them. 
```
docker run -it --rm --name bob_agent --expose 8030-8033 --expose 5678 -p 8030-8033:8030-8033 -p 5678:5678 -e "DOCKERHOST=192.168.65.3" -e "LEDGER_URL=http://192.168.65.3:9000" -e "TAILS_SERVER_URL=http://192.168.65.3:6543" acapy-main-agent-backchannel -p 8030
```

Moopli:
@nodlesh I'm not sure why it fails on your end, the agents start fine (and pass most of their steps too) when I do the following:
```
$ [my local von-network dir]/manage start
$ [my local indy-tails-server dir]/docker/manage start
$ LEDGER_URL_CONFIG="http://localhost:9000" TAILS_SERVER_URL_CONFIG="http://localhost:6543" ./manage run -d acapy-main -t @AcceptanceTest -t @AIP20 -t ~@wip
```
What, if anything, do you see in the agent logs?

nodlesh:
This is Bob not starting properly.  
```
Starting aca-py agent ...
From ledger_url: http://192.168.65.3:9000/genesis
Listening to backchannel on port 9030
Listening to web_hooks on port 9033
Matched operation: {'rfc': '0000 General Backchannel Commands', 'command': 'X', 'response': '', 'topic': 'status', 'method': 'GET', 'operation': '', 'id': '', 'data': '', 'description': 'Return agent status, 200 if active or 418 otherwise', 'id_details': '', 'data_details': '', 'result_details': 'status'}
Matched operation: {'rfc': '0000 General Backchannel Commands', 'command': 'X', 'response': '', 'topic': 'status', 'method': 'GET', 'operation': '', 'id': '', 'data': '', 'description': 'Return agent status, 200 if active or 418 otherwise', 'id_details': '', 'data_details': '', 'result_details': 'status'}
.
.
.
```
It is missing `Got DID: ` and onward.

ianco:
`LEDGER_URL_CONFIG="http://localhost:9000" TAILS_SERVER_URL_CONFIG="http://localhost:6543" ./manage run ...`

The agents are not going to be able to connect to `localhost:9000` nor `localhost:6543`, they need an IP that is recognized inside the docker network.

On the other hand, when the `manage` script checks for the von-network and ledger, it needs to use `localhost`

ianco:
With the latest updates, when I run:  `./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip`, everything behaves as expected (ledger & tails server start & then shut down at the end of the test)

When I run:  `LEDGER_URL_CONFIG=http://localhost:8000 TAILS_SERVER_URL_CONFIG=http://localhost:6543 ./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip` (running the ledger & tails server manually) the `manage` script doesn't detect the ledger and then shuts down.

I tried to substitute `192.168.65.3` (the docker IP address on my machine) for `localhost` and no luck.

swcurran:
So we need a special case for the scenario of running von-network and indy-tails-server locally on docker, since they have two different URLs depending on how they are accessed (within bash/.manage and within docker).  And presumably, we don't want to start and stop the servers on every run, as that will slow down development.

How about a special flag (e.g. "--local" or the like) to say that those two services are running locally?  We're going to have this with every service that runs in a docker container.  For example, I would imagine we'll see this if we start to have a "did:orb" service running within a docker container.

ianco:
> So we need a special case for the scenario of running von-network and indy-tails-server locally on docker, since they have two different URLs depending on how they are accessed (within bash/.manage and within docker). And presumably, we don't want to start and stop the servers on every run, as that will slow down development.
> 
> How about a special flag (e.g. "--local" or the like) to say that those two services are running locally? We're going to have this with every service that runs in a docker container. For example, I would imagine we'll see this if we start to have a "did:orb" service running within a docker container.

:+1: on `--local`, except prob `--local-ledger` and `--local-tails` separate parameters

Moopli:
@ianco I'm not sure why `LEDGER_URL_CONFIG=http://localhost:8000 TAILS_SERVER_URL_CONFIG=http://localhost:6543 ./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip` is failing, it should be using the [localhost url](https://github.com/hyperledger/aries-agent-test-harness/pull/244/files#diff-181229424893bb65d94a74c2132b8b9e5adfe851464fdb5cb9f49e8a8204be7bR19) for the [ledger liveness check](https://github.com/hyperledger/aries-agent-test-harness/pull/244/files#diff-181229424893bb65d94a74c2132b8b9e5adfe851464fdb5cb9f49e8a8204be7bR262) and the [internal URL](https://github.com/hyperledger/aries-agent-test-harness/pull/244/files#diff-181229424893bb65d94a74c2132b8b9e5adfe851464fdb5cb9f49e8a8204be7bR25) for the [agent containers](https://github.com/hyperledger/aries-agent-test-harness/pull/244/files#diff-181229424893bb65d94a74c2132b8b9e5adfe851464fdb5cb9f49e8a8204be7bR380).

Moopli:
@ianco btw are you sure you're running von-network on localhost:**8000**, not localhost:**9000**?

ianco:
> @ianco btw are you sure you're running von-network on localhost:**8000**, not localhost:**9000**?

Aha good catch, let me check and try again

OK confirm this works when I use the correct ports:

`LEDGER_URL_CONFIG=http://localhost:9000 TAILS_SERVER_URL_CONFIG=http://localhost:6543 ./manage run -d afgo-interop -b acapy-main -v 20 -t @RFC0023 -t ~@wip`

ianco:
With this command:  `LEDGER_URL_CONFIG=http://localhost:9000 TAILS_SERVER_URL_CONFIG=http://localhost:6543 ./manage run -d acapy-main -v 20 -t @RFC0023 -t ~@wip`

(i.e. using aca-py agents for everything)

...  Bob agent doesn't start up.  I suspect there is a port getting grabbed somewhere.  aca-py opens up a range of ports (at least 3 off the top of my head) so possibly conflicts with something being used by one of the new services?

ianco (changes requested):
Looks good but possible port conflict?

troyronda:
> For example, I would imagine we'll see this if we start to have a "did:orb" service running within a docker container.

@swcurran We plan to add a "did:orb" service in a followup.

Moopli:
@ianco the only place I can find a port in the 9030-9039 range being mapped, outside bob-agent, is the demo, where that range is assigned to faber. I think it must be a different failure, I'll push a tweak in a moment to allow logs to be fetched from crashed agents so you can see what the problem is.

ianco:
The aca-py agents are randomly hanging up when trying to register a public DID on the ledger.  The agent gets to the point of calling the endpoint to register their DID, but then the call never returns (or at least there's no evidence in the log files).

The DID's however *ARE* getting registered on the ledger - when I inspect the ledger browser after the tests have completed, *all* the agents have DID's registered, even the ones that failed to start.

I notice there is some changes to how the docker networking is configured, maybe that's a cause (?)   I don't see any reason why one aca-py should register successfully and others randomly fail.

troyronda:
@Moopli @ianco Any next steps on getting this situation resolved and this PR merged?

Moopli:
note, just rebased on master ^

nodlesh:
This is now working for me after recloning everything. Good to go!

ianco (approved):

ianco:
> @Moopli @ianco Any next steps on getting this situation resolved and this PR merged?

Everything looks good, I'll leave it for @nodlesh to merge.

PR #266, opened 2021-06-23
Cred id bypass invite create

nodlesh:
This PR updates the secondary get of the inviter's connection_id at the point the invitee accepts the invitation in the 160 Connection Protocol tests. There is now a call to get the connection id based on the invitation id.


swcurran (approved):
LGTM

PR #273, opened 2021-07-01
Back-channels cleanup

andrewwhitehead:
This PR applies Black formatting to the back channel implementations, fixes a bunch of Python warnings and removes unused imports. It also removes the checks for the indy-sdk Python library which were inherited from the demo agents, but aren't necessary here (if it's needed but not installed, then the tests will fail anyway).

It also fixes a failure when running `manage stop` on my system by adjusting the check for `docker_result`, and aborts `manage build` if a docker build fails.

### DIFF
--- a/aries-backchannels/acapy/acapy_backchannel.py
+++ b/aries-backchannels/acapy/acapy_backchannel.py
@@ -3,28 +3,35 @@ import functools
 import json
 import logging
 import os
-import random
 import subprocess
 import sys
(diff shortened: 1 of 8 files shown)


ianco (approved):

ianco:
Looks good to me, I'll leave it for @nodlesh to merge in case it causes him any conflicts

nodlesh:
👍

PR #274, opened 2021-07-02
Switch to 'slim' docker image for acapy back channel

andrewwhitehead:
This reduces the image size by about 400Mb on my system. It should also be a little faster, combining the pip package installs and removing one apt-get update.


ianco (approved):

swcurran (approved):
Ran full test under current and this branch and they matched perfectly.

PR #276, opened 2021-07-02
Auto response state handling

### LINKED ISSUE #268: Auto Response State Handling
As an Aries Framework Interop Tester I want to test my auto responding agent in the AATH, so that auto responding agents can be verified for interop along side other auto responding agents or non-auto responding agents with the same tests. 

**Problem:** The tests as they stand check transitional protocol states as stated in the RFC for the connection protocols (and other protocols as well). With auto responding agents (like Verity), the requests are responded to automatically and the agent protocol states transition to a further state in the process than the tests expect them to be in. IE. When the invitee sends a connection request, the inviter may auto respond. The test expects the state to be `requested` but both agents have already moved onto `responded`. This causes the connection test to fail, however both agents have been successful thus far. 

**Solutions (Proposals):**

1. No transitional state checks in connection tests. It may be overkill for the AATH to be checking all of the interim states in the protocols. Though the RFCs state these are the states after the given operations, the overarching goal of the AATH is to facilitate interop testing between agents of differing capabilities. In this case, kicking off the protocol, loosely driving the intermediate operations, and then checking that the final state of the protocol is correct, should suffice to accomplish the AATH goal. 

2. There is some value in checking the intermediate states as it relates to failing the test at the appropriate time when an error happens, which would be useful in backchannel development and debugging. Also capturing the expressed description in the RFCs. If we want to keep this benefit it is proposed to amend a response from the backchannel to the test harness with `transitioned_states`. The `state` still represents the current state of the agent in the protocol, but if the state is not expected from a non-auto respond perspective, the test harness can check the transitioned_states and determine that we are working with an auto responding agent, pass the step if the transitioned states include the current expected state and make decisions on the remaining steps to run. 
```
{
   "state":"responded",
   "transitioned_states":[
      "invited",
      "requested"
   ]
}
```

3. Require the backchannel to implement a webhook to return each transitional state as it comes in.

**Chosen Solution:**
The community seems to be leaning toward option 1. Given that, the proposal is to create a new version of the connection test step definitions that do not check intermediate protocol states. It will only check for 200 response codes to carry on. These new step definition will supplant the old ones once we are sure we are not negatively impacting other agents with existing backchannels already running these tests.

nodlesh:
This PR removes the intermediate state handling in the Connection and DID Exchange Protocols. The tests still check for an initial state (ie invited) and and end state of `completed`. As for auto responding and auto accepting, the backchannel is going to have to know it's in that auto mode and reply back to the calls it doesn't need to make to the agent with 200 response code.  Backchannel developers can see the examples in the DID Exchange and Connection POST Handlers in the Aca-py Backchannel.

Closes #268 

@smithbk @Matt-Spence If there are any other state asserts that get in the way, let me know and I'll remove them as well.  Issue Cred and Proof will follow the same pattern as above but will come later.


swcurran:
I think this is OK, but I found this odd.  I ran using this PR and Master and got the same results (good), but saw this errors -- are these expected?

```
Command Line: ./manage run -d acapy -t @RFC0160,@T002-RFC0023,@T005-RFC0023 -t ~@wip

Failing scenarios:
  features/0023-did-exchange.feature:69  Establish a connection with DID Exchange between two agents with an implicit invitation
  features/0160-connection.feature:72  Inviter Sends invitation for one agent second agent tries after connection
  features/0160-connection.feature:89  Inviter Sends invitation for one agent second agent tries during first share phase
  features/0160-connection.feature:102  Inviter Sends invitation for multiple agents
  features/0160-connection.feature:119  Establish a connection between two agents who already have a connection initiated from invitee
  features/0160-connection.feature:145  Establish a connection between two agents but gets a request not accepted report problem message -- @1.1 
  features/0160-connection.feature:146  Establish a connection between two agents but gets a request not accepted report problem message -- @1.2 

0 features passed, 2 failed, 5 skipped
4 scenarios passed, 7 failed, 90 skipped
63 steps passed, 7 failed, 843 skipped, 0 undefined
Took 0m48.752s
```

Also -- will this break existing Backchannels (Aries Framework .NET and Aries Framework JavaScript) tests?

Should we wait to here from Matt/Keith or go ahead with merging this?

nodlesh:
Yes, that is odd and shouldn't be failing. I ran with ` -t @AcceptanceTest -t @RFC0023,@RFC0160 -t ~@wip ` but that excluded a couple of exception tests you have failing. I'll have to check that before merge.

These changes should have no effect on the other backchannels. Dotnet and afgo are running fine as per the runset in the daily gha. Just ran javascript, seems to be some issues there but not sure if it is becasue of this PR. Have to investigate.

nodlesh:
@swcurran could you try your execution again? I cannot make the same tests fail. In your run command you exclude @wip, however, you have failures in your run for @wip tests. Strange. 

These three tests are failing for me, inside and outside this PR. Two are expected; 72 and 89. The scenario at 119 has been failing for a while now and I'm still investigating if it is expected.  
```
features/0160-connection.feature:72  Inviter Sends invitation for one agent second agent tries after connection
  features/0160-connection.feature:89  Inviter Sends invitation for one agent second agent tries during first share phase
  features/0160-connection.feature:119  Establish a connection between two agents who already have a connection initiated from invitee
```

swcurran (approved):
Re-ran the tests -- full (~@wip) on main and this branch and they matched.  w00t!

PR #298, opened 2021-07-19
chore: update afj backchannel

### LINKED ISSUE #282: Javascript Agent Randomly failing to receive message from Aca-py on connection accept-invitation
When running any test with Javascript as Acme and Aca-py as Bob, the establishment of a connection is not guarenteed. Randomly (probably a timing issue), when Aca-py send a `connection/accept-invitation`, the Javascript agent throws this error:
```
[2021-07-08T18:57:44.611] [INFO ] [TSED] - received agent message {}
(node:29) UnhandledPromiseRejectionWarning: IndyError: CommonInvalidParam3
    at Object.callback (/src/node_modules/indy-sdk/src/wrapIndyCallback.js:15:10)
(node:29) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:29) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
``` 

I am unsure of why this happens and why it doesn't seem to error out at other runs. The test that I can get it to happen almost all the time when debugging in VSCode is T001-RFC0036. Running the tests with the manage script at the command line may or may not produce the same problem at the same point in the test suite. 

Not sure if it has something to do with the merge of #255? The day after this merge is when the problem started to appear. 

@TimoGlastra maybe you can get a chance to take a look at what might be going on here?

TimoGlastra:
Updates AFJ backchannel to latest release on NPM. Haven't updated yet to use the main branch but that should be a minor addition now.

Let's hope this also resolves #282 (works for me locally)


manage
```diff
@@ -755,11 +755,11 @@ pushd ${SCRIPT_HOME} >/dev/null
 
 case "${COMMAND}" in
   build)
-      buildImages ${@}
+      buildImages ${@} --platform linux/amd64
     ;;
 
   rebuild)
-      buildImages --no-cache ${@}
+      buildImages --no-cache ${@} --platform linux/amd64
```
TimoGlastra, line 762: This is needed to run on M1 mac

swcurran (approved):
Ran main and this branch and found this one had fewer errors -- an additional 3 features and 26 steps passed.  Given that the error was intermittent, that might not be a final answer, but it bodes well.

PR #299, opened 2021-07-20
Mobile wait states

ianco:
Add a delay to wait for mobile to make a connection request before responding


ianco:
OK this "fixes" the issue with running the mobile backchannel (timing around when the inviter can accept the invitee's request).

It works with Trinsic with no issues.

With esatus the connection gets established and the subsequent credential and proof work, however esatus eventually thinks the connection is not valid and drops it.

Would be interested in seeing results from others ...

nodlesh (approved):
LGTM - no negative side effects on the regular acapy tests.

Issue #259, opened 2021-06-16
Simplify/Change Proof v1 test data to just the proposal and request


nodlesh:
The existing data passed is
```
 "presentation_proposal": {
                "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/request-presentation",
                "comment": "This is a comment for the request for presentation.",
                "request_presentations~attach": {
                    "@id": "libindy-request-presentation-0",
                    "mime-type": "application/json",
                    "data":  data
                }
            }
```
and data is
```
data = {
                    "requested_attributes": {
                        "attr_1": {
                            "name": "attr_1",
                            "restrictions": [
                                {
                                    "schema_name": "test_schema." + context.issuer_name,
                                    "schema_version": "1.0.0"
                                }
                            ]
                        }
                    }
                }
```

we really just need the 
```
"proof_request": {
...
}
```
or 
```
"presentation_proposal": {
...
}
```

TimoGlastra:
While we're at it. I found it a bit confusing that in both v1 and v2 of the present proof protocols the send-request endpoints expect an object containing a `presentation_proposal` instead of `presentation_request`

nodlesh:
@TimoGlastra Yes, agreed.

TimoGlastra:
Also: The `send-proposal` method takes a presentation preview, but with `requested_attributes` and `requested_predicates` instead of `attributes` and `predicates`. Everything else is the same

TimoGlastra:
So this:
```
    presentation_proposal = {
        "presentation_proposal": {
            "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/presentation-preview",
            "comment": "This is a comment for the presentation proposal.",
            "requested_attributes": requested_attributes,
            "requested_predicates": requested_predicates
        }
    }
```

should actually be:
```
    presentation_proposal = {
        "presentation_proposal": {
            "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/present-proof/1.0/presentation-preview",
            "comment": "This is a comment for the presentation proposal.",
            "attributes": requested_attributes,
            "predicates": requested_predicates
        }
    }
```

PR #300, opened 2021-07-21
fix: afj and dotnet interop issues

TimoGlastra:
AFJ had a lot of changes that were not updated in AATH. 4 issues:

1. AFJ sends https messages by default. Although .NET supports https, the backchannel did not -> Updated .NET backchannel to support https messages
2. AFJ includes both the `IndyAgent` and `did-communication` services in a did doc. .NET errors if other services than `IndyAgent` are present. As I don't suspect AF.NET to be updated in the near future I'v made a PR to AFJ: https://github.com/hyperledger/aries-framework-javascript/pull/402. We will wait with sending this until we support did exchange.
3. .NET crashes if no `mime-type` is present ([optional according to the RFC](https://github.com/hyperledger/aries-rfcs/blob/master/features/0036-issue-credential/README.md#mime-type-and-value)). AFJ doesn't send a mime-type if it is not explicitly set -> Updated AFJ to emit `mime-type` value of `text/plain` if not set
4. AFJ didn't take the presentation preview property quirk (as described [here](https://github.com/hyperledger/aries-agent-test-harness/issues/259#issuecomment-884553477)) into account. -> fixed.

This PR should be merged after https://github.com/hyperledger/aries-framework-javascript/pull/402 is merged and I had the chance to update the dependency in here.

afj-dotnet and acapy-dotnet-javascript fully passing now

### DIFF
--- a/.github/workflows/test-harness-acapy-afj.yml
+++ b/.github/workflows/test-harness-acapy-afj.yml
@@ -11,9 +11,9 @@ name: test-harness-acapy-javascript
 # notably those that involve revocation.
 #
 # Current
-# 
+#
 # Most of the tests are running. The tests not passing are being investigated.
-# 
+#
 # *Status Note Updated: 2021.03.18*
 #
 # End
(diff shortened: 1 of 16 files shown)


TimoGlastra:
Ready to merge! This should solve I think (almost?) all tests where AFJ is involved

nodlesh (approved):
👍

PR #307, opened 2021-07-25
Adding details about drilling into the Allure results for a failing test

swcurran:

### DIFF
--- a/docs/aries-interop-intro.md
+++ b/docs/aries-interop-intro.md
@@ -81,7 +81,7 @@ The Allure reports accessible from this site provide a lot of information about 
    4. If it has passed in earlier runs, drill into the passing test to see if you can learn anything from that.
    5. Dive into the weeds...look at the stack trace, find the associated code, and see if you can figure out what happened. Find anything?  Report it via an issue, or even better, submit a Pull Request to fix the issue. Remember to consider all of the possible places an error could occur (above).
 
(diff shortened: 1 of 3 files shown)


docs/aries-interop-intro.md
```diff
@@ -19,69 +19,87 @@ What unites Aries agents are the standards and protocols they aim to adhere to,
...
+   3. Click on the "History" tab to see if it has always failed, or if this is a new issue.
+   4. If it has passed in earlier runs, drill into the passing test to see if you can learn anything from that.
+   5. Dive into the weeds...look at the stack trace, find the associated code, and see if you can figure out what happened. Find anything?  Report it via an issue, or even better, submit a Pull Request to fix the issue. Remember to consider all of the possible places an error could occur (above).
+
+In addition to drilling into a specific test scenario (aka "stories")/case (aka "behavior")/step, you can look at the entire runset history. On the left side menu, click on "Overview", and then take a look at the big "history" graph in the top right, showing how the runset execution has varied over time. Ideally, it's all green, but since you started from a runset that had failures, it won't be. Pretty much every part of the overview page is a drill down link into more and more detailed information about the runset, a specific run of the runset, a specific test case and so on. Lots to look at!
```
nodlesh: "entire runset history" is not actually the case. It is capped at 20 executions. This was to manage space on the Allure server. We can increase that. We thought 20 was good enough to show some recent history, and we probably don't care about any older than that.

nodlesh (dismissed):
👍

swcurran:
I'll adjust -- thanks for the info.

swcurran:
As I wrote that, I thought that we should check on much storage we are using, so capping it at 20 is a very good idea :-).

nodlesh:
Yes agreed. We could move off the historical records to another server, but it may have minimal value. Right now it is deleted by Allure.

nodlesh (approved):

PR #315, opened 2021-08-05
chore: update to work with latest afgo (removes a bug workaround)

Moopli:


nodlesh (approved):
Ran the existing afgo tests. Looks good.

PR #316, opened 2021-08-09
fix for failing mobile tests with new connection reuse PR

nodlesh:
This PR will fix the failing Mobile Tests because of the interjection of getting a public did on the `And "Faber" and "Bob" have an existing connection` step. The handler now just accepts the 500 not found from the mobile backchannel and continues to create the connection without reporting the 500. 


swcurran (approved):
LGTM -- worked with Trinsic

Issue #326, opened 2021-09-03
Dotnet backchannel errors for Acapy-Dotnet runset when All interim state checks are removed. 


nodlesh:
The AATH no longer checks for interim states in protocol progression. In the Issue Credential and Proof tests the Dotnet Agent/Backchannel fails when running with Acapy with these state checks removed. Running with dotnet playing all roles does not produce the errors. There is a call `expected_agent_state()` in the tests where not calling it in two cases, causes the failures. 

**HOW TO REPRODUCE**
Comment out the following lines of code.
`@when('"{issuer}" offers a credential')` on line 254 of `0036-issue-credential.py`.
```
assert expected_agent_state(context.holder_url, "issue-credential", context.cred_thread_id, "offer-received")
```
`@when('"{verifier}" sends a request for proof presentation to "{prover}"')` on line 174 of `0037-present-proof.py`
```
assert expected_agent_state(context.prover_url, "proof", context.presentation_thread_id, "request-received")
```
Run the following commands
```
./manage rebuild -a dotnet-master -a acapy-main
./manage run -d acapy-main -b dotnet-master -t @AcceptanceTest -t ~@wip -t ~@ProofProposal -t ~@RFC0023 -t ~@DIDExchangeConnection
```

**RESULTS**
```
  features/0036-issue-credential.feature:9  Issue a credential with the Holder beginning with a proposal
  features/0036-issue-credential.feature:37  Issue a credential with the Holder beginning with a proposal with negotiation
  features/0036-issue-credential.feature:53  Issue a credential with the Issuer beginning with an offer
  features/0036-issue-credential.feature:66  Issue a credential with the Issuer beginning with an offer with negotiation
  features/0037-present-proof.feature:19  Present Proof where the prover does not propose a presentation of the proof and is acknowledged -- @1.1 
  features/0037-present-proof.feature:20  Present Proof where the prover does not propose a presentation of the proof and is acknowledged -- @1.2 
  features/0037-present-proof.feature:55  Present Proof of specific types and proof is acknowledged with a Drivers License credential type -- @1.1 
  features/0037-present-proof.feature:56  Present Proof of specific types and proof is acknowledged with a Drivers License credential type -- @1.2 
  features/0037-present-proof.feature:74  Present Proof of specific types and proof is acknowledged with a Biological Indicators credential type -- @1.1 
[...]

Issue #325, opened 2021-09-02
Auto Responding version of Acapy GHA Test Runset


nodlesh:
With the support of Auto Responding in Issue Credential and Present Proof in AATH, it may be advisable to create 1-n daily git hub action runsets that have all or a combination of auto-respond flags enabled when running acapy.  

To facilitate this, there are two approaches that have been discussed. More alternatives are welcome. 
1. Make a new image for acapy names something like `acapy-auto`. So executing the tests with auto enabled would look like this `./manage build -a acapy-auto` and `./manage run -d acapy-auto -t ...`
2. Add the ability to pass a configuration file to the build command. `./manage build -a acapy-main -config AcapyAutoEnabled`

swcurran:
I don't think the auto-config settings should be relevant for the test suite. At most it should be buried in the backchannel.  I'm going to close this. Please reopen if you think we still need to do this.

PR #327, opened 2021-09-03
Auto respond handling for Issue Credential/Proof 

### LINKED ISSUE #320: Auto Response State Handling in Issue Credential/Proof V1 & 2
This PR follows the work done in Auto response state handling #276, and is related to #268.

This will remove the intermediate state handling in the Issue Credential and Proof Protocols V1 and V2. The tests still check for an initial state and and end state. As for auto responding and auto accepting, the backchannel is going to have to know it's in that auto mode and reply back to the calls it doesn't need to make to the agent with 200 response code. Backchannel developers can see the examples in the DID Exchange and Connection POST Handlers in the Aca-py Backchannel.

nodlesh:
This PR adds support for auto-respond in AATH, along with the removal of interim agent state checks.  These changes affect Issue Credential 1.0 and 2.0 along with Presentation 1.0 and 2.0.  

There are two instances (one for Issue Credential and one for Presentation) where the state check could not  be removed at this time. Removing them causes problems with the Acapy to Dotnet runset. 
One is in `@when('"{issuer}" offers a credential')` on line 254 of `0036-issue-credential.py`.
```
assert expected_agent_state(context.holder_url, "issue-credential", context.cred_thread_id, "offer-received")
```
The other is in `@when('"{verifier}" sends a request for proof presentation to "{prover}"')` on line 174 of `0037-present-proof.py`
```
assert expected_agent_state(context.prover_url, "proof", context.presentation_thread_id, "request-received")
```
For backchannel developers that do not need this state check in an auto responding agent, please just return 200, with the expected state for now. These lines will be removed once the problem is determined and fixed. 
See outstanding issues below for further details of this issue. 

**HOW TO VERIFY**
This PR affects all run sets and all agents. No need to run them all, but a few runset executions would be prudent. Get the runset configurations from [here](https://github.com/hyperledger/aries-agent-test-harness/tree/main/.github/workflows). 

**OUTSTANDING ISSUES**
Mobile Issue: @ianco I've added the state checks in certain instances where the mobile tests are expected to wait, to the acapy backchannel. However, I only get 9 out of 10 passing. At the end of some of the tests I also get a `something went wrong` error in Trinsic Wallet, but the credential is issued properly and the presentation is made. Wondering if you can take a look and let me know if it is something I've done or this is expected? 

Dotnet Backchannel Issue: @TimoGlastra It seems, with all the state checks removed, Dotnet or the combination of Dotnet with Acapy has a problem. Executing Dotnet in all agent rolls doesn't cause this issue. I've logged an issue for this #326 . I'm wondering if you can take a look and help diagnose the problem. I'm assuming the Dotnet backchannel was doing something to advance the protocol in those state checks. 

**AWARENESS**
@smithbk @Matt-Spence 

**SPAWNED ISSUES**
#325 
#326 
#330 

Closes #320

### DIFF
--- a/aries-backchannels/acapy/acapy_backchannel.py
+++ b/aries-backchannels/acapy/acapy_backchannel.py
@@ -208,13 +208,13 @@ class AcaPyAgentBackchannel(AgentBackchannel):
             ("--label", self.label),
             # "--auto-ping-connection",
             # "--auto-accept-invites",
-            "--auto-accept-requests",
-            "--auto-respond-messages",
-            "--auto-respond-credential-proposal",
-            "--auto-respond-credential-offer",
-            "--auto-respond-credential-request",
-            "--auto-respond-presentation-proposal",
-            "--auto-respond-presentation-request",
+            #"--auto-accept-requests",
+            #"--auto-respond-messages",
+            #"--auto-respond-credential-proposal",
+            #"--auto-respond-credential-offer",
+            #"--auto-respond-credential-request",
+            #"--auto-respond-presentation-proposal",
+            #"--auto-respond-presentation-request",
             ("--inbound-transport", "http", "0.0.0.0", str(self.http_port)),
             ("--outbound-transport", "http"),
             ("--admin", "0.0.0.0", str(self.admin_port)),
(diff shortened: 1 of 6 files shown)


ianco (approved):

swcurran (approved):
LGTM

swcurran:
FYI — I think the PR tests are failing because they are trying to push to Allure, which they probably shouldn’t be doing.  I think if we dropped that, all would work.

PR #328, opened 2021-09-03
Pin tar version for javascript backchannel

ianco:

Pin the tar version due to security vulnerability.

Manually "fixed" the yarn.lock file, not sure that this is the best approach :-S


nodlesh (approved):
👍

PR #335, opened 2021-09-13
Acapy-b-verity test runset

nodlesh:

This PR adds an acapy-b-verity runset for daily execution in GHA. 
A acapy-b-verity project has been added to allure. 
The aries interop reports should be automatically updated upon test completion.

### DIFF
--- a/.github/workflows/test-harness-acapy-verity.yml
+++ b/.github/workflows/test-harness-acapy-verity.yml
@@ -1,0 +1,53 @@
+name: test-harness-acapy-verity
+# RUNSET_NAME: "ACA-PY to Verity"
+# Scope: AIP 1.0
+# Exceptions: None
+#
+# Summary
+#
+# This runset uses the current main branch of ACA-Py for all of the agents except Bob (holder),
+# which uses the dev branch of Evernym Verity. The runset covers all of the AIP 1.0 tests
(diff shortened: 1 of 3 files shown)


swcurran (approved):
LGTM - we'll adjust if things don't work quite right

swcurran:
The tests ran, and the gen-interop script worked correctly.  Neither of the two tests ran, so that needs to be investigated:

```
Tags:  --tags=@RFC0160 --tags=@AcceptanceTest --tags=~@wip

features/0160-connection.feature  FF

Failing scenarios:
  features/0160-connection.feature:21  establish a connection between two agents -- @1.1 
  features/0160-connection.feature:46  Connection established between two agents but inviter sends next message to establish full connection state -- @1.1 

0 features passed, 1 failed, 6 skipped
0 scenarios passed, 2 failed, 96 skipped
4 steps passed, 2 failed, 881 skipped, 0 undefined
Took 0m0.092s

```

As well, in the comments from the "runset" GHAction, there are comments that get pulled into the interop page. In this case, the runset status says "All the Connection tests are running. All tests are passing", which is not accurate. 

We should probably do a pass of all of the GHActions for those comments and update them to be accurate.  Or we could remove them entirely, but I think that would lose an subjective assessment of the status of the runset.

PR #337, opened 2021-09-14
Remove PR triggered tests until Allure problem can be resolved

swcurran:

Removing the triggering of the tests on PR addition.  The tests are failing on every run because they don't have permission to run the Allure publishing. To re-add:

- figure out and correct the Allure publishing issue.
- make a new action that runs the test, but doesn't do the Allure testing.

For now, we'll just turn off the test runs.

### DIFF
--- a/.github/workflows/test-harness-acapy-afgo.yml
+++ b/.github/workflows/test-harness-acapy-afgo.yml
@@ -18,9 +18,6 @@ name: test-harness-acapy-afgo
 #
 # End
 on:
-  pull_request:
-    branches:
-    - main
   workflow_dispatch:
   schedule:
     - cron: "5 0 * * *"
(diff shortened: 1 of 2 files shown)


swcurran:
@nodlesh -- can you please review/approve this one?  I'd like to stop merging PRs with failed tests.

Thanks!

nodlesh:
Sure. These tests are not failing actually. It is the push to allure which needs GitHub secrets to be able to push. We didn't intend to put the secrets in forks. If we want these to run in forks then we can create new actions that do not push to allure, or we can find some sort of condition we can check in the existing actions.  For right now, I'll remove these executions from the PRs.

swcurran:
Agreed. This PR does that -- removes them.  So if you approve this PR, we'll be good to go.

PR #338, opened 2021-09-16
added sleeps in the javascript backchannel to wait for appropriate states to proceed

nodlesh:
This PR should fix the Javascript with acapy daily test runset failures.  It problems ended up being the removal of the interim state checks in the tests, since we are not checking (and waiting) for the proper state before proceeding, the waits have to be done in the javascript backchannel instead.  These are all passing on my local machine, if some still fail in the daily runs, then we may have to increase the new wait times.  

Use the following command to test this:
`./manage run -d javascript -b acapy-main -t @AcceptanceTest -t ~@wip -t @AIP10`


swcurran (approved):
Ran the tests and all executed correctly -- nice!

PR #341, opened 2021-09-17
Add backchannel for findy-agent

lauravuo-techlab:
Signed-off-by: Laura Vuorenoja <laura.vuorenoja@op.fi>

Add backchannel for testing compatibility with [Findy Agency](https://findy-network.github.io) agent. Current interoperability target for findy-agent is IOP1.0 excluding revocation.

### DIFF
--- a/.github/workflows/test-harness-acapy-findy.yml
+++ b/.github/workflows/test-harness-acapy-findy.yml
@@ -1,0 +1,55 @@
+name: test-harness-acapy-findy
+# RUNSET_NAME: "ACA-PY to findy"
+# Scope: AIP 1.0
+# Exceptions: Revocation
+#
+# Summary
+#
+# This runset uses the current main branch of ACA-Py for all of the agents except Bob (holder),
+# which uses the latest release of findy-agent. The runset covers all of the AIP 1.0 tests that 
+# are known to work with the findy-agent as the holder. Excluded are those tests that involve Revocation.
+#
+# Current
+# 
+# All of the tests being executed in this runset are passing.
+# 
+# *Status Note Updated: 2021.09.14*
+#
+# End
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "20 1 * * *"
+defaults:
+  run:
+    shell: bash
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
+    env:
+      LEDGER_URL_CONFIG: "http://localhost:9000"
+      TAILS_SERVER_URL_CONFIG: "http://localhost:6543"
+    steps:
+      - name: checkout-test-harness
+        uses: actions/checkout@v2
+        with:
+          path: test-harness
+      - name: run-von-network
+        uses: ./test-harness/actions/run-von-network
+      - name: run-indy-tails-server
+        uses: ./test-harness/actions/run-indy-tails-server
+      - name: run-test-harness-wo-reports
+        uses: ./test-harness/actions/run-test-harness-wo-reports
+        with:
+          BUILD_AGENTS: "-a acapy-main -a findy"
+          TEST_AGENTS: "-d acapy-main -b findy"
+          TEST_SCOPE: "-t @AcceptanceTest -t @AIP10 -t ~@wip -t ~@revocation"
+          REPORT_PROJECT: acapy-b-findy
+        continue-on-error: true
+      - name: run-send-gen-test-results-secure
+        uses: ./test-harness/actions/run-send-gen-test-results-secure
+        with:
+          REPORT_PROJECT: acapy-b-findy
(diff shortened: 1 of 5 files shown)


swcurran (approved):
Awesome job.  I was going to confirm the tests that should be passing, but I see you have a runset ready to go -- excellent.  I ran the Findy Test Agent running all tests and assumed what you are saying -- AIP 1.0 and no revocation.  Perfect.

Would love to have any feedback you have about the AATH -- things we could improve?

nodlesh:
3 allure projects have been created here, https://allure.vonx.io/allure-docker-service-ui/
`findy`
`findy-b-acapy`
`acapy-b-findy`

swcurran:
Please add a findy-afj-dotnet project as well (a/m, b, f respectively).

nodlesh:
So an allure project named `findy-b-afj-f-dotnet`?

swcurran:
Sure

nodlesh:
Done

lauravuo-techlab:
@swcurran 

Thanks for reviewing this so promptly!

I think AATH is an excellent tool for testing interoperability and even functionality in general. We integrated the tests also to our findy-agent release pipeline – agent release is not happening if the AATH tests are not passing, so we can easily catch e.g. regression with our client APIs.

This was my first time when I used openapi specification based code generation – I think it also worked rather well and eased the developer job when starting with a new backchannel from scratch.

Some improvements that came to mind along the way:
* It would be excellent if the test script would support running agent and backchannel in different containers. We have a microservice architecture and running agent needs both the core service (findy-agent) and the auth service (findy-agent-auth). Now I need to bundle the three services (core, auth, backchannel) into a single AATH-dedicated image. Instead, I’d rather use image per service as we have the service images already built in our Docker registry. There would be no need to build the images for the test run, just pull them from the registry. This model would also ensure that the tests are run with exactly the same image as which the deployment is using (not only the same version of the codes).
* As we are using AATH also for regression testing, I was wondering if there is a parameter or switch for the manage-script that would stop the execution on failure and return an error code? Currently I am just checking that the test output contains some magic string to confirm that the test was successful, but it would be better to 1) save time if tests fail 2) fail CI pipeline automatically (shell cmd to return error)
* We aim to build the findy-agent client API such that it has minimal amount of direct indy-related features. The target is to hide the underlying technology from the client, so that we can add other credential formats in the future without the need to refactor the API too much. One thing in the AATH/acapy test logic was against this principle – the need to define attribute ids in proof requests. I guess this shouldn’t be necessary – the proof request attribute ids is really something that indy-sdk requires, but the client shouldn’t be bothered. However, a small thing and we added the the support for setting the id, so no actual obstacles there.

swcurran:
Interesting comments -- thanks for sharing them.  I'll start with the first.

The "one container" approach was to make a consistent interface for of the Test Agents (where TA is the Backchannel+Whatever is being Tested). It allows us to just have a single way to start each Agent (E.g. Acme or Bob) without worrying about any details of the specific implementation.  We did give some consideration that each TA could be a docker-compose file, and the "./manage" script could assemble the docker-compose pieces at runtime to a complete docker-compose file to be run...but we think that's even worse.   While we know the intention of a container is to put one thing in it to aid in management in real environments, it's not a requirement, especially in a test environment like this.  I do agree that it feels odd to put everything into one container.

We could have the backchannel in one container, and the "thing being tested" in another, but I don't think that helps you, as you would have to put "everything else" into the one container.  It's really not too different.

So, I'd say we're not likely to pursue the idea of multi-container TAs unless someone has a clever, manageable way to do this. If you have an idea for that, let us know and we'll rethink this.

swcurran:
I think adding a parameter to "stop on error" would be a good idea and happy to see that in the backlog.  Since our default approach is to run all tests and always push the pass/fail results to Allure (test results repository), a fail status on the test run would mean the results don't get to Allure.  However, if we aren't running with Allure, an error status would make sense.

Happy to see that in the backlog.

swcurran:
It would be good to get a little more detail on the third issue.  I think I get it, but if you could highlight a specific example, that would be helpful.  In general, I agree with hiding details that are not needed in the backchannel API makes sense.

lauravuo-techlab:
Thanks @swcurran! I will create issues for our findings with more elaborate descriptions. What comes to the first bullet, I don't have any better ideas (yet). It's a shame we cannot run docker-compose in a docker container 😄

WadeBarnes:
> It's a shame we cannot run docker-compose in a docker container

You can, here is an example of a docker-compose container being built and then used to build another image using docker-compose in the container.  https://github.com/hyperledger/indy-test-automation/blob/main/system/docker/prepare.sh#L68-L85

WadeBarnes:
It would require some work to AATH, but I think it would be possible to define a multi-container TA in a docker-compose file and then have AATH start that up in a docker-in-docker image that has docker-compose installed (as in the above example).  Both would use a defined network (also in the example above) allowing AATH to communicate with the backchannel which in-turn would communicate with the other TA components.  Then each TA could be defined using it's own docker-compose file and container/network topology.

PR #352, opened 2021-09-28
Findy afj dotnet runsets

nodlesh:
This PR adds runsets for findy, dotnet, and afj combinations. All tests are failing because of connection issues . My recommendation is to put them up failing, and have people who are familiar with these agents diagnose/fix the issue.


swcurran (dismissed):

nodlesh:
No update from findy. Would suggest putting it up failing.

nodlesh:
@swcurran  A lot more tests passing after a findy update today. yml files have been updated with the latest results. The only things failing now are when findy plays the holder. That is being investigated, but this can be merged as is.

swcurran (approved):

PR #353, opened 2021-09-29
feat: afgo-acapy present proof interop

Moopli:


nodlesh (approved):
This is excellent. Ran all afgo runsets and everything passed. Also ran a couple of other agent runsets just to be sure no side effects from the manage script edits.

swcurran:
Awesome -- that's great!

PR #359, opened 2021-10-07
Add support for testing RFC-0044 mime types

Moopli:
- Adds `/agent/start` backchannel POST command, which tells the backchannel to (re)start its agent, and includes a set of parameters to include in the agent's configuration (whether as command-line parameters, environment variables, or otherwise). Parameters have generic names on the AATH side, and must be translated by the backchannel for the given agent implementation. In this PR only one parameter is introduced, `mime-type` to specify the default mime type for the agent's outgoing messages as per Aries RFC 0044.
- Adds support for `/agent/start` to afgo backchannel, handling the `mime-type` parameter.
- Adds a new step that (re)starts a given agent with a given set of extra parameters.
- Adds a BDD feature for RFC 0044, that restarts Acme and Bob with given default mime types, and performs didexchange to demonstrate permissive agent acceptance of incoming mime types.

### DIFF
--- a/.github/workflows/test-harness-afgo.yml
+++ b/.github/workflows/test-harness-afgo.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           BUILD_AGENTS: "-a afgo-master"
           TEST_AGENTS: "-d afgo-master"
-          TEST_SCOPE: "-t @RFC0023,@T001.1-RFC0036,@RFC0453,@RFC0454 -t ~@wip -t ~@CredFormat_Indy"
+          TEST_SCOPE: "-t @RFC0023,@RFC0044,@T001.1-RFC0036,@RFC0453,@RFC0454 -t ~@wip -t ~@CredFormat_Indy"
           REPORT_PROJECT: afgo 
         continue-on-error: true
       - name: run-send-gen-test-results-secure
--- a/aries-backchannels/afgo/afgo_backchannel.py
+++ b/aries-backchannels/afgo/afgo_backchannel.py
@@ -71,6 +71,8 @@ class AfGoAgentBackchannel(AgentBackchannel):
             params
         )
 
+        self.output_handler_futures = []
+
         # Afgo : RFC
         self.connectionResponderStateTranslationDict = {
             "invited": "invitation-received",
@@ -179,6 +181,11 @@ class AfGoAgentBackchannel(AgentBackchannel):
             "issue-credential-v2": "/issuecredential/",
             "proof": "/presentproof/",
             "proof-v2": "/presentproof/"
+        }
+
+        # Generic AATH parameter to AFGO commandline parameter
+        self.map_cmdline_params = {
+            "mime-type": "media-type-profiles"
         }
 
     def get_agent_args(self):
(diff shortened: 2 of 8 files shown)


Moopli:
TODO:
- [x] Add new backchannel command to OpenAPI spec

swcurran:
Super clever idea!  I like this.  @ianco -- what do you think of this?  As long as the start parameters are based on RFCs, this is pretty powerful -- especially for testing Community Update RFCs.

ianco:
> Super clever idea! I like this. @ianco -- what do you think of this? As long as the start parameters are based on RFCs, this is pretty powerful -- especially for testing Community Update RFCs.

I like it!

FYI in the aca-py integration tests, we've taken a similar approach (starting/stopping agents for each test scenario, with different aca-py features such as multi-tenancy, aip 1 vs 2, etc), vs the approach in AATH to run all tests on the same agent instance.  This is a nice feature where we can restart the AATH agent-under-test if necessary.

swcurran:
Ran it for `afgo-master` and it worked perfectly.  Tried it for `afgo-interop` (command: `./manage run -d afgo-interop -t @RFC0044`) and it didn't work (which is probably OK), but it looks like something minor, since it gets through all of the steps.  Should that work?

Here is the output:

``` bash
@T002-RFC0044
  Scenario Outline: Perform DID Exchange between two permissive agents that have different default envelope MIME types -- @1.1   # features/0044-mime-types.feature:40
    Given we have "2" agents                                                                                                     # features/steps/0160-connection.py:16 0.000s
      | name | role      |
      | Acme | requester |
      | Bob  | responder |
    And "Acme" is running with parameters "{"mime-type":"didcomm/aip1"}"                                                         # features/steps/0044-mime-types.py:12 1.027s
    And "Bob" is running with parameters "{"mime-type":"didcomm/v2"}"                                                            # features/steps/0044-mime-types.py:12 1.029s
    When "Bob" sends an explicit invitation                                                                                      # features/steps/0023-did-exchange.py:18 20.053s
    And "Acme" receives the invitation                                                                                           # features/steps/0023-did-exchange.py:180 0.047s
    And "Acme" sends the request to "Bob"                                                                                        # features/steps/0023-did-exchange.py:210 3.088s
    And "Bob" receives the request                                                                                               # features/steps/0023-did-exchange.py:221 0.507s
    And "Bob" sends a response to "Acme"                                                                                         # features/steps/0023-did-exchange.py:263 0.027s
      Assertion Failed: resp_status 400 is not 200; {"code":2005,"message":"did exchange client - accept exchange request: failed to execute state action 'responded': outboundDispatcher.Send: failed to pack msg: packMessage: failed to pack: authcrypt Pack: failed to encrypt payload: jweencryptWithSender: failed to wrap 
```

Moopli:
@swcurran
> it didn't work (which is probably OK), but it looks like something minor, since it gets through all of the steps. Should that work?

No, didexchange in general isn't expected to work afgo-interop to afgo-interop :P
WIth the various bits of custom logic that allow acapy interop right now, some deeper changes would need to be done to afgo-interop to allow it to work with itself.

swcurran (approved):
Looks great!

swcurran:
@andrewwhitehead -- are we close to being able to use this with ACA-Py?  @ianco -- does it make sense for you to look at adding this feature to the ACA-Py Backchannel?

PR #366, opened 2021-10-15
add a full suite of Media Type Profile negotiation tests

Moopli:
Tests demonstrate OOB media-type-profile negotiation across different scenarios, including success and failure.


swcurran (approved):
Looks good -- ran the afgo-master tests for @RFC0044

PR #367, opened 2021-10-15
fix(LE CA issues): fixing Let's Encrypt CA expiration fallout

pruneau628:
- updating root CA for verity backchannel build
- adding a warning in manage when a mobile agent is used

### DIFF
--- a/manage
+++ b/manage
@@ -20,7 +20,7 @@ LEDGER_URL_HOST="${LEDGER_URL_CONFIG:-http://localhost:9000}"
 LEDGER_URL_INTERNAL="${LEDGER_URL_CONFIG:-http://${DOCKERHOST}:9000}"
 TAILS_SERVER_URL_HOST="${TAILS_SERVER_URL_CONFIG:-http://localhost:6543}"
 TAILS_SERVER_URL_INTERNAL="${TAILS_SERVER_URL_CONFIG:-http://${DOCKERHOST}:6543}"
-CLOUD_AGENCY_URL="${CLOUD_AGENCY_URL:-http://localhost:8070}"
+CLOUD_AGENCY_URL="${CLOUD_AGENCY_URL:-http://localhost:8000}"
 # important: inside the internal URLs, we replace "://localhost:" with "://${DOCKERHOST}:"
 #   so it works inside docker.
 LEDGER_URL_INTERNAL="$(echo ${LEDGER_URL_INTERNAL} | sed "s/:\/\/localhost:/:\/\/${DOCKERHOST}:/" )"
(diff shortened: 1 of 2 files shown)


swcurran (commented):
Looks great!  Thanks.  Before I approve I wanted to check on the changes to the "Cloud Agency" stuff.  Should those be there?

I was also going to mention the "but it might print twice", but I see you were clever about that.  Nice!

Let me know about the cloud agency - related changes.

manage
```diff
@@ -20,7 +20,7 @@ LEDGER_URL_HOST="${LEDGER_URL_CONFIG:-http://localhost:9000}"
 LEDGER_URL_INTERNAL="${LEDGER_URL_CONFIG:-http://${DOCKERHOST}:9000}"
 TAILS_SERVER_URL_HOST="${TAILS_SERVER_URL_CONFIG:-http://localhost:6543}"
 TAILS_SERVER_URL_INTERNAL="${TAILS_SERVER_URL_CONFIG:-http://${DOCKERHOST}:6543}"
-CLOUD_AGENCY_URL="${CLOUD_AGENCY_URL:-http://localhost:8000}"
+CLOUD_AGENCY_URL="${CLOUD_AGENCY_URL:-http://localhost:8070}"
```
nodlesh: We changed the port of the cloud agency because of a port conflict with our Orb service. Is there a reason to have this set back to 8070?

manage
```diff
@@ -710,14 +710,11 @@ startHarness(){
 
   startServices auto
 
-  # Only start cloud agency if vcx is in the agent list to run.
-  if [[ "${ACME}" = "aries-vcx" ]] || [[ "${BOB}" = "aries-vcx" ]] || [[ "${MALLORY}" = "aries-vcx" ]] || [[ "${FABER}" = "aries-vcx" ]]; then
-    startCloudAgency
+  startCloudAgency
```
nodlesh: Maybe you haven't merged with main? You are missing some conditional statements to starting the Cloud Agency.

manage
```diff
@@ -743,10 +740,7 @@ startHarness(){
 
   echo
   # Allure Reports environment.properties file handling
-  # Only do this if reporting parameter is passed. 
-  if [[ "${REPORT}" = "allure" ]]; then
-    writeEnvProperties
-  fi
+  writeEnvProperties
```
nodlesh: Need the latest from main.

nodlesh (dismissed):

manage
pruneau628: Yes sorry, this is my first pull request, and I did all this quickly to have it out before late Friday.
Since my fork was literally minutes old when I submitted, I'm not sure why those change where not based on the latest from main.
They sure need to be reverted, I'll do it this week-end if I have the chance, or next week at the latest.

swcurran (approved):
All good.

I didn't get the @Mobile trigger to fire, but I think that is OK.  And once this is no longer an issue, this can be removed.

Thanks for doing this!

PR #371, opened 2021-10-24
chore: update afj, remove setTimeouts

### LINKED ISSUE #251: Update AFJ backchannel to run on main branch
Currently the AFJ backchannel has a packaged version of AFJ, meaning it is not testing against the latest code.

There were quite some changes in the framework since last updated so probably also needs some updates to the backchannel code.

- [ ] Update the AFJ backchannel to work with the current code in the main branch of AFJ
- [ ] Update `Dockerfile.javascript` to use the currently released version of AFJ on NPM
- [ ] Add `Dockerfile.javascript-main` to clone and use the main branch of AFJ (see ACA-Py implementation)
- [ ] Update the github action workflows to use the `Dockerfile.javascript-main` instead of `Dockerfile.javascript`

TimoGlastra:

- Updates AFJ to latest version (and also uses `latest` as version indicator). This means if a new version is released it should be picked up by the CI. This also means the AFJ backchannel will probably break more often, as we're still making a lot of breaking changes. I hope this will make sure I keep it up to date :)
- Removes the `setTimeouts` from the controller and instead uses events to know when we can take the next action (@nodlesh)
- @lauravuo-techlab it seems that updating AFJ fixed the issue with findy (yay! :)). On my local machine the following command (`./manage run -d javascript -b findy -t @AcceptanceTest -t @AIP10 -t ~@wip -t ~@revocation`) all tests succeeds

### DIFF
--- a/aries-backchannels/javascript/Dockerfile.javascript
+++ b/aries-backchannels/javascript/Dockerfile.javascript
@@ -46,7 +46,8 @@ RUN yarn install
 
 # Copy other depedencies
 COPY javascript/server .
+COPY javascript/ngrok-wait.sh ./ngrok-wait.sh
 
 # For now we use ts-node. Compiling with typescript
 # doesn't work because indy-sdk types are not exported
-ENTRYPOINT [ "yarn", "ts-node", "src/index.ts" ]
+ENTRYPOINT [ "bash", "ngrok-wait.sh"]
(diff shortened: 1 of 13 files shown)


lauravuo-techlab:
Thanks @TimoGlastra for the investigation! I can confirm that tests are passing with this change 💪

swcurran (approved):
Approved based on comments from @lauravuo-techlab 

Thanks, Timo!

PR #387, opened 2021-11-22
Fix timeout logic on wait for thread status

ianco:


swcurran:
Any idea why this step:   

`And "Acme" and "Bob" have an existing connection                      # features/steps/0160-connection.py:273 10.459s`

Takes 10.459 seconds to complete?  Seems really long and seems to be part of every test, so it really slows things down.

swcurran (approved):
Ran `main` and this PR branch and both worked identically.  Approving.

Interestingly, it took 14 seconds longer to run this branch.  Hmmm... :-)

PR #390, opened 2021-11-23
Shorten sleep during connections

ianco:

### DIFF
--- a/aries-test-harness/features/steps/0160-connection.py
+++ b/aries-test-harness/features/steps/0160-connection.py
@@ -188,8 +188,9 @@ def step_impl(context, invitee, inviter):
 @when('"{inviter}" receives the connection request')
 @given('"{inviter}" receives the connection request')
 def step_impl(context, inviter):
-    inviter_url = context.config.userdata.get(inviter)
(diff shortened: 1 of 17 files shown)


aries-test-harness/features/steps/0160-connection.py
```diff
@@ -372,7 +372,7 @@ def step_impl(context, sender, receiver):
     sender_connection_id = context.connection_id_dict[sender][receiver]
 
     data = {"comment": "acknowledgement from " + sender}
-    sleep(5)
+    sleep(1)
```
swcurran: Is it possible to use the polling loop in these case as we mentioned yesterday -- `wait_until(timeout, polling_interval)`, so in this case `wait_until(20,0.5)`?

swcurran:
Agree that we probably shouldn't adjust the sleep.  In running it ACA-Py to ACA-Py it only reduced the time to 6 seconds.

I'm more worried about the use of sleep() at all, and if it is possible for us to switch from sleep() (which we all know will never be right) to polling everywhere. That will enable events to be processed as fast as they happen vs. we wait for what we hope is long enough -- usually waiting way too long in most cases, and too short in others.

PR #393, opened 2021-11-25
fix(afj): send proof request fixes and prettier formatting

TimoGlastra:


swcurran (approved):
From 0 tests passing to all of them.  Nice!

PR #398, opened 2021-12-06
Add support for changes to internal Docker networking.

WadeBarnes:
- Networking changes introduced in Docker 4.1.x and forward on Windows and MAC stop the direct use of the internal docker host IP returned by the `docker run --rm --net=host eclipse/che-ip` process.  On Windows and MAC `host.docker.internal` needs to be used for internal connections between containers.


WadeBarnes:
The changes in this PR are also dependent on the corresponding changes to `von-network` here; https://github.com/bcgov/von-network/pull/188

Users will have to delete the cached copy of `von-network` that gets cloned by AATH.

nodlesh (approved):
👍

PR #399, opened 2021-12-08
Centralize docker host IP script.

WadeBarnes:
- This is an update to the previous commit to update the support for Docker networking.  The `getDockerHost` has been pulled out and placed in a central location where it can be updated as needed rather than having to update dozens of scripts the next time there is a change.

### DIFF
--- a/manage
+++ b/manage
@@ -5,22 +5,9 @@
 export VALID_AGENTS=$(ls aries-backchannels/*/Dockerfile.* | sed "s/^.*file.//" |  tr "\n" " " | sort -u)
 
 export MSYS_NO_PATHCONV=1
-# ====================================================================
-# Set the DOCKERHOST address depending on host system
-# --------------------------------------------------------------------
-function getDockerHost() {
-  (
-    unset dockerHostAddress
-    if [[ $(uname) == "Linux" ]] ; then
-      dockerHostAddress=$(docker run --rm --net=host eclipse/che-ip)
-    else
-      dockerHostAddress=host.docker.internal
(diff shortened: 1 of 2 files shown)


ianco (approved):
very cool

nodlesh (approved):
😎👍

swcurran (commented):
I'm not sure that a BCGov repo is the best place for this.  At minimum, but Aries things, it should be in Aries, and for BCGov things in BC Gov.

What about continuing the eclipse-che process -- e.g. having a container that spits back the right thing?  However, use a container that we have better insight into and can change when needed.  E.g. that detects that the host is Mac and/or Windows and returns `host.docker.internal`?

WadeBarnes:
> I'm not sure that a BCGov repo is the best place for this. At minimum, but Aries things, it should be in Aries, and for BCGov things in BC Gov.
> 
> What about continuing the eclipse-che process -- e.g. having a container that spits back the right thing? However, use a container that we have better insight into and can change when needed. E.g. that detects that the host is Mac and/or Windows and returns `host.docker.internal`?

I put the script into the BCGov repo because most of the scripts I needed to update (30 projects) were in BCGov.  As for the container process, I never really liked it much, it's a bit heavy weight for what it does.  Having to maintain a new container would require additional overhead and have the same "where should it live" problem to solve.  I also see the issue with the Linux containers being fixed in the future making the eclipse-che process or any container process we create completely obsolete.

Always open for discussion.

swcurran (approved):

PR #401, opened 2021-12-11
aries-vcx: Send presentation ack always

mirgee:
acapy now seems to expect presentation ack even if the presentation is evaluated as invalid. This fixes some currently failing tests.

Signed-off-by: Miroslav Kovar <miroslavkovar@protonmail.com>

### DIFF
--- a/aries-backchannels/aries-vcx/src/controllers/presentation.rs
+++ b/aries-backchannels/aries-vcx/src/controllers/presentation.rs
@@ -187,11 +187,11 @@ impl Agent {
         if !vec![VerifierState::Finished, VerifierState::Failed].contains(&verifier.get_state()) {
             return Err(HarnessError::from_msg(HarnessErrorType::ProtocolError, "Presentation not received"));
         }
+        verifier.send_ack(&connection.send_message_closure()?)?;
         self.dbs.verifier.set(&verifier.get_thread_id()?, &verifier)?;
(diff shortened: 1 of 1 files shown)


swcurran:
That sounds like a bug in ACA-Py -- need to look into that. Did you catch what test that had this behaviour?

mirgee:
@swcurran You can see the failing tests [here](https://allure.vonx.io/api/allure-docker-service/projects/aries-vcx-b-acapy/reports/latest/index.html?redirect=false#behaviors/42f7406febc7b666c8bfd79d3ad464f1/429d9127f903be29/).

mirgee:
@swcurran Also (unrelated), during the preparation of the previous PR that acapy changed the thread decorators on connection request so that `pthid` matches the `@id` of the invitation and `thid` matches the `@id` of the connection request. I think it makes sense in case of a "public invitation" because there may be replies from several different entities and thus new branching threads. Either way, it does not seem clearly specified in the [connection RFC](https://github.com/hyperledger/aries-rfcs/tree/main/features/0160-connection-protocol) (it should be, IMO). I just wanted make sure - was this change intentional?

mirgee:
@swcurran These issues still remain relevant. I applied this change in #422.

swcurran:
@andrewwhitehead -- can you please take a look at this issue.  What is the right behaviour based on the RFC?

Thanks for following up @mirgee.

PR #412, opened 2022-01-19
fix: sleep is not defined

TimoGlastra:

Fixes an issue with `sleep` not being defined resulting in all present proof tests to fail (not sure how I missed this, as the tests ran successfully yesterday...)


nodlesh (approved):
👍

PR #414, opened 2022-01-20
further separated crons for less parallel execution

nodlesh:

This may fix the cancelled jobs in the daily interop test runs. If this does work, that means ultimately we need to move to the matrix model so this doesn't happen again. This current submission is just to prove the crons are/were the problem.

### DIFF
--- a/.github/workflows/test-harness-acapy-ariesvcx.yml
+++ b/.github/workflows/test-harness-acapy-ariesvcx.yml
@@ -9,7 +9,7 @@ name: test-harness-acapy-ariesvcx
 on:
   workflow_dispatch:
   schedule:
-    - cron: "50 1 * * *"
+    - cron: "50 2 * * *"
 defaults:
   run:
     shell: bash
--- a/.github/workflows/test-harness-findy-acapy.yml
+++ b/.github/workflows/test-harness-findy-acapy.yml
@@ -19,7 +19,7 @@ name: test-harness-findy-acapy
 on:
   workflow_dispatch:
   schedule:
-    - cron: "25 1 * * *"
+    - cron: "25 2 * * *"
 defaults:
   run:
     shell: bash


swcurran:
How about changing all the jobs to run on different days of the week?

Change them - use the last argument of cron: cron: `30 5 * * 2,4` -- set the last arg (day of week) of half the jobs to `1,3,5` and the other half to `2,4,6`.  The runners can rest on Sunday (7).

swcurran:
I'll merge this one for now, but I'm not sure this will give us enough info to know what the problem is.

swcurran (approved):

nodlesh:
We absolutely can run on different days. The matrix approach might be more elegant, however more complex.

swcurran:
@nodlesh -- Looking at failing and passing instances of the same job (examples below), it looks like perhaps there is a hang up sometimes between the completion of the behave run and the start of the cleanup.  Is there anything going on in that part of the process that might be a problem?  Note the termination of the job after the behave completed, but before the "Cleanup" message was printed.

### Failing

This job was cancelled after 46m, including building and running the entire test suite (which took 26m):

```
features/0036-issue-credential.feature  ....
features/0037-present-proof.feature  .............
features/0160-connection.feature  ...
features/0183-revocation.feature  ..................
4 features passed, 0 failed, 4 skipped
38 scenarios passed, 0 failed, 72 skipped
343 steps passed, 0 failed, 652 skipped, 0 undefined
Took 25m49.296s
Error: The operation was canceled.
```

### Passing 

A successful run took 39m (although the entire run was over 1h according to the summary):

```
features/0036-issue-credential.feature  ....
features/0037-present-proof.feature  .............
features/0160-connection.feature  ...
features/0183-revocation.feature  ..................
4 features passed, 0 failed, 4 skipped
38 scenarios passed, 0 failed, 72 skipped
343 steps passed, 0 failed, 652 skipped, 0 undefined
Took 23m18.541s

Cleanup:
  - Shutting down all the agents ...
Done
stopping local uniresolver...
Stopping uni-resolver-web.local ... 
Stopping driver-did-sov.local   ... 
Stopping driver-did-sov.local   ... done
Stopping uni-resolver-web.local ... done
Found orphan containers (mysql, agency) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Removing uni-resolver-web.local ... 
Removing driver-did-sov.local   ... 
Removing driver-did-sov.local   ... done
Removing uni-resolver-web.local ... done
Network aath_network is external, skipping
```

swcurran:
What's the matrix model?

nodlesh:
I'm not seeing that pattern you illustrate repeated. When I look at test-harness-acapy-ariesvcx
Passing
```
features/0036-issue-credential.feature  ....
features/0037-present-proof.feature  .............
features/0160-connection.feature  ...
features/0183-revocation.feature  ..................
4 features passed, 0 failed, 4 skipped
38 scenarios passed, 0 failed, 72 skipped
343 steps passed, 0 failed, 652 skipped, 0 undefined
Took 23m18.541s
```
Failing
```
features/0036-issue-credential.feature  ....
features/0037-present-proof.feature  FFFFFFFFFFFFF
features/0160-connection.feature  ...
features/0036-issue-credential.feature  ....
features/0037-present-proof.feature  .............
features/0160-connection.feature  ...
```
It is failing in the middle of connection and revocation tests. Not even getting to any final test results. 

I'm not sure if the matrix strategy will be a solution or not, I've just read it may be a solution if you have cancelled jobs from parallel operations. I haven't dug very deep to be honest.

nodlesh:
It is also the case that test-harness-findy-acapy has a successful run, a report generated, and uploaded to Allure, but then, for some reason, gets started 4 or 5 more times later. Some of the later runs get canceled. I wanted to move those two runsets out of the execution window of the other runsets to see if they still have the same behaviour.

TimoGlastra:
What is the issue with the scenarios running in parallel? Seems like if we can fix that issue, we don't have to push out the cron job times

nodlesh:
Well according to yesterdays run, it isn't because of parallel execution. Moving those two runsets out of the window of the rest of the runsets made no difference. They have the exact same problem, so it is something particular to those runsets. They look identical to the other runsets, and they work when running locally. They both pass on the first run, then they are triggered again and get cancelled. If we can figure out why they are triggered for multiple runs, that may be all we need to fix.

PR #416, opened 2022-01-24
added the case not to copy logs if in GITHUB ACTIONS

nodlesh:

To further diagnose the cancel on the acapy-aries-vcx run, this PR makes copying agent logs conditional on being in GITHUB_ACTIONS. Copying the logs is the next thing that happens at about the point the cancel happens.

### DIFF
--- a/manage
+++ b/manage
@@ -787,17 +787,21 @@ runTests() {
   rm ${BEHAVE_INI_TMP}
 
   # Export agent logs
-  mkdir -p .logs
-  docker logs acme_agent > .logs/acme_agent.log
-  docker logs bob_agent > .logs/bob_agent.log
-  docker logs faber_agent > .logs/faber_agent.log
-  docker logs mallory_agent > .logs/mallory_agent.log
+  if [[ "${GITHUB_ACTIONS}" = "true" ]]; then
+    echo "Exporting Agent logs."
+    mkdir -p .logs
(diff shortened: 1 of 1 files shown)


ianco (approved):

swcurran (approved):

manage
```diff
@@ -787,17 +787,21 @@ runTests() {
...
-  docker logs bob_agent > .logs/bob_agent.log
-  docker logs faber_agent > .logs/faber_agent.log
-  docker logs mallory_agent > .logs/mallory_agent.log
-
-  if [[ "${USE_NGROK}" = "true" ]]; then
-    docker logs acme_agent-ngrok > .logs/acme_agent-ngrok.log
-    docker logs bob_agent-ngrok > .logs/bob_agent-ngrok.log
-    docker logs faber_agent-ngrok > .logs/faber_agent-ngrok.log
-    docker logs mallory_agent-ngrok > .logs/mallory_agent-ngrok.log
+  if [[ "${GITHUB_ACTIONS}" = "true" ]]; then
```
TimoGlastra, line 790: How I read this is that it will only copy the logs if `GITHUB_ACTIONS == true`? Like the stament of `USE_NGROK`. 

Shouldn't this be inverted?

nodlesh: Yes, yes it should...

PR #419, opened 2022-01-25
docs(api): issue credential and present proof v2

TimoGlastra:
Adds open api definitions for the issue credential and present proof v2 endpoints


swcurran (approved):
LGTM - I think.  This will evolve as needed.

PR #436, opened 2022-02-15
fix(afj): revocation tests

TimoGlastra:
Fixes broken AFJ revocation tests. Also adds the missing @RFC0441 tags that we now use to only run the revocation tests following RFC0441


swcurran (approved):
LGTM

PR #444, opened 2022-03-03
feat: didcomm v2 connection steps and afgo support

Moopli:

### DIFF
--- a/aries-backchannels/afgo/afgo_backchannel.py
+++ b/aries-backchannels/afgo/afgo_backchannel.py
@@ -572,6 +572,10 @@ class AfGoAgentBackchannel(AgentBackchannel):
             (resp_status, resp_text) = await self.handle_agent_POST(command)
             return (resp_status, resp_text)
 
+        elif command.topic == "oob-v2":
+            (resp_status, resp_text) = await self.handle_oob_v2_POST(command)
+            return (resp_status, resp_text)
+
         return (501, "501: Not Implemented\n\n")
 
     async def handle_agent_POST(self, command: BackchannelCommand):
@@ -620,6 +624,80 @@ class AfGoAgentBackchannel(AgentBackchannel):
         await self.start_process_with_extra_args(args=args)
 
         return (200, "{}")
+
+    async def handle_oob_v2_POST(self, command: BackchannelCommand):
+        operation = command.operation
+        data = command.data
+
+        if operation == "create-invitation":
+            (
+                orb_did_status,
+                orb_did_text,
+                priv_key_kids,
+            ) = await self.get_orb_did()
+            if orb_did_status != 200:
+                err_msg = f"Couldn't retrieve my public did: status={orb_did_status} body={orb_did_text}"
+                print(err_msg)
+
+                return (500, err_msg)
+
+            my_pub_did = json.loads(orb_did_text)["did"]
+
+            afgo_endpoint = "/outofband/2.0/create-invitation"
+            create_req = {
+                "from": my_pub_did
+            }
+
+            if data and "goal-code" in data:
+                create_req["body"] = {
+                    "goal-code": data["goal-code"]
+                }
+
+            (resp_status, resp_text) = await self.admin_POST(afgo_endpoint, create_req)
+
+            return (resp_status, resp_text)
+
+        if operation == "accept-invitation":
+            data["my_label"] = "dummy-label"
+
+            afgo_endpoint = "/outofband/2.0/accept-invitation"
+
+            (resp_status, resp_text) = await self.admin_POST(afgo_endpoint, data)
+            # afgo's reponse object looks like {"connection_id":"<connection ID>"}
+            return (resp_status, resp_text)
+
+        return (501, f"didcomm v2 operation '{operation}' not implemented")
+
+    async def handle_oob_v2_GET(self, command: BackchannelCommand):
+        operation = command.operation
+
+        if operation == "invitation-connection":
+            connection_id = ""
+            invitation_id = command.record_id
+
+            afgo_endpoint = f"/connections"
+
+            (resp_status, resp_text) = await self.admin_GET(afgo_endpoint)
+            if resp_status != 200:
+                return (resp_status, f"failed to query afgo connections, response: {resp_text}")
+
+            resp_json = json.loads(resp_text)
+            if "results" not in resp_json:
+                return (500, f"failed to query afgo connections, missing connection list from {resp_text}")
+
+            for connection in resp_json["results"]:
+                log_msg(f"Check connection: {connection}")
+                if connection["InvitationID"] == invitation_id:
+                    connection_id = connection["ConnectionID"]
+                    break
+
+            response = {
+                "connection_id": connection_id
+            }
+
+            return (200, json.dumps(response))
+
+        return (501, f"didcomm v2 operation '{operation}' not implemented")
 
     async def handle_out_of_band_POST(self, command: BackchannelCommand):
         operation = command.operation
@@ -821,15 +899,18 @@ class AfGoAgentBackchannel(AgentBackchannel):
         record_id = command.record_id
         data = command.data
 
-        if record_id is None:
-            agent_operation = f"{self.TopicTranslationDict[topic]}{self.issueCredentialOperationTranslationDict[operation]}"
-        else:
-            agent_operation = (
+        def make_agent_operation(record_id: str):
+            return (
                 self.TopicTranslationDict[topic]
                 + record_id
                 + "/"
                 + self.issueCredentialWRecIDOperationTranslationDict[operation]
             )
+
+        if record_id is None:
+            agent_operation = f"{self.TopicTranslationDict[topic]}{self.issueCredentialOperationTranslationDict[operation]}"
+        else:
+            agent_operation = make_agent_operation(record_id)
 
         if data is not None:
             log_msg(
@@ -904,6 +985,10 @@ class AfGoAgentBackchannel(AgentBackchannel):
                 ]
                 their_did = issue_credential_actions_msg["message"]["Properties"][
                     "theirDID"
+                ]
+
+                record_id = issue_credential_actions_msg["message"]["Properties"][
+                    "piid"
                 ]
 
             else:
@@ -1025,6 +1110,10 @@ class AfGoAgentBackchannel(AgentBackchannel):
 
                     data.pop("filter")
 
+            if record_id is not None:
+                # in some cases we overwrite record_id with the correct piid
+                agent_operation = make_agent_operation(record_id)
+
             log_msg(
                 f"Data translated by backchannel to send to agent for operation: {agent_operation}",
                 data,
@@ -1064,6 +1153,15 @@ class AfGoAgentBackchannel(AgentBackchannel):
             return (resp_status, resp_text)
 
         elif operation == "send-request":
+            (wh_status, wh_text) = await self.request_response_issue_credential(
+                rec_id=record_id, message_name="issue-credential-actions-msg"
+            )
+            issue_credential_actions_msg = json.loads(wh_text)
+
+            record_id = issue_credential_actions_msg["message"]["Properties"]["piid"]
+
+            agent_operation = make_agent_operation(record_id)
+
             log_msg(
                 f"Data translated by backchannel to send to agent for operation: {agent_operation}",
                 data,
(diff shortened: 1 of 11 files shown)


aries-test-harness/features/didcomm-v2.feature
```diff
@@ -0,0 +1,48 @@
...
+      And "Acme" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Bob" is running with parameters "{"mime-type":["didcomm/v2"]}"
+
+      When "Acme" creates a didcomm v2 invitation
+      And "Bob" accepts the didcomm v2 invitation from "Acme"
+
+      # Bob must send a message to acme
+      When "Bob" requests mediation from "Acme"
+
+      And "Acme" has received a didcomm v2 message from "Bob" and created a connection
```
Moopli: Preferably, would have a step after this, to verify that they both have a connection. The standard step ("Acme" and "Bob" have a connection) expects a protocol-complete webhook message, for either didexchange or connection-protocol, and didcomm v2 doesn't work that way.

DRK3 (dismissed):

TimoGlastra (commented):
This is great @Moopli!

Two questions:
1. I see this is using RFC 0453 while creating a connection using DIDComm v2. Is this creating a DIDComm v2 connection while using DIDComm v1 message structure for the issue credential protocol? Or is this actually using the DIDComm v2 variant of the issue credential protocol (https://github.com/decentralized-identity/waci-presentation-exchange/blob/main/issue_credential/README.md). If the latter it would be good to not use the term `RFC0453`, but rather give it a new tag. We don't have RFCs anymore with didcomm v2 so maybe something like `@ICV3`. Or `IssueCredentialV3` to be more explicit?
2. I see you're using the `didcomm-v2` topic for the new endpoints. If we start moving more towards didcomm v2 I think the topic can get quite bloated. In the past we've mostly seperated topics by feature or protocol. Maybe we can make it `oob-v2`?

Moopli:
@TimoGlastra The afgo implementation of issue-credential V2 and V3 use the same protocol logic, and same controller API, with the internal message structure being translated into didcomm-v1-style issue-credential V2 or didcomm-v2-style issue-credential v3 messages based on whether the connection used is a didcomm v1 or didcomm v2 connection - so it made sense for us to reuse the RFC0453 steps and backchannel commands,  to minimize how much we had to add to the test harness. 

1. Maybe we can figure out a new API that minimizes the number of new steps/commands we need to add? For example, by adding new tags as you suggest, and adjusting steps and before-scenario handlers to take them into account?
2. That's the same reason I put the new endpoints under the `didcomm-v2` topic - if issue-credential and present-proof can use the same steps for V2 and V3, then only the didcomm-v2-specific functionality needs its own steps, and so far that's only connection creation with oob v2.

DRK3 (dismissed):

aries-test-harness/features/didcomm-v2.feature
nodlesh: Could you not just create a new step phrased a little differently and implemented it in didcimm-v2.py? 
like 
`"Acme" and "Bob" have a didcomm-v2 connection`
as you did with `When "Acme" creates a didcomm v2 invitation`

aries-test-harness/features/didcomm-v2.feature
```diff
@@ -0,0 +1,48 @@
+@DIDComm-V2 @UsesCustomParameters
```
nodlesh: Since there is no RFC, can we have a list of links (commented out) that specify the functionality that these tests are derived from?

aries-test-harness/features/didcomm-v2.feature
```diff
@@ -0,0 +1,48 @@
...
+Feature: DIDComm V2 Establishing Connections (with Issue Credential example)
```
nodlesh: I'm really struggling with this merging of protocols in one feature. This feature file is going to get pretty big if it eventually holds all DIDcomm v2 tests. I am really leaning toward Issue Credential V3 feature file, and connection v3?

nodlesh:
I agree with @TimoGlastra in point 1, RFC0453 doesn't have any place in the tag list. No tests here were derived from that RFC. Same goes for @RFC0593. You can reuse steps from RFC0453 without classifying the tests as that RFC. 
On Point 2 I also agree. The operation topics need to align with the protocol as the pattern that had been set. 
issue-credential-v3 and oob-v2 sounds like better alignment. 

That same logic should apply to the feature files as well I think. So an issue-credential-v3.feature and oob-v2.feature

Are there really no RFCs, there is nothing out there that can uniquely identify these features? No specification referenced with some sort of identifier?

aries-test-harness/features/didcomm-v2.feature
```diff
@@ -0,0 +1,48 @@
...
+         | Bob  | holder |
+      And "Acme" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Bob" is running with parameters "{"mime-type":["didcomm/v2"]}"
+
+      And "Acme" is ready to issue a "json-ld" credential
+
+      When "Acme" creates a didcomm v2 invitation
+      And "Bob" accepts the didcomm v2 invitation from "Acme"
+
+      When "Bob" proposes a "json-ld" credential to "Acme" with <credential_data>
```
nodlesh: Should be an `And` here.

aries-test-harness/features/didcomm-v2.feature
```diff
@@ -0,0 +1,48 @@
...
+      When "Acme" creates a didcomm v2 invitation
+      And "Bob" accepts the didcomm v2 invitation from "Acme"
+
+      # Bob must send a message to acme
+      When "Bob" requests mediation from "Acme"
+
+      And "Acme" has received a didcomm v2 message from "Bob" and created a connection
+
+   @T001.1-RFC0453-DIDComm-V2 @RFC0453 @AIP20 @RFC0593 @critical @AcceptanceTest @DIDExchangeConnection @CredFormat_JSON-LD @Schema_DriversLicense_v2 @ProofType_Ed25519Signature2018 @DidMethod_key
+   Scenario Outline: Issue a JSON-LD Ed25519Signature2018 credential with the Holder beginning with a proposal, using DIDComm v2 for the connection
```
nodlesh: If the point of this test is to issue a cred with two actors with a didcimm-v2 connection, I think we can shorted this up a bit. The When clause should always be the actions taken directly related to the point of the test. So it could be slightly simplified like this, 
```
Given "2" agents
...
And "Acme" is running with parameters "{"mime-type":["didcomm/v2"]}"
And "Bob" is running with parameters "{"mime-type":["didcomm/v2"]}"
And "Acme" is ready to issue a "json-ld" credential
And "Acme and "Bob" have a didcimm v2 connection 
When "Bob" proposes a "json-ld" credential to "Acme" with <credential_data>
And "Acme" offers the "json-ld" credential
And "Bob" requests the "json-ld" credential
And "Acme" issues the "json-ld" credential
And "Bob" acknowledges the "json-ld" credential issue
Then "Bob" has the "json-ld" credential issued
```

aries-test-harness/features/steps/didcomm-v2.py
```diff
@@ -0,0 +1,77 @@
...
+    assert resp_status == 200, f"resp_status {resp_status} is not 200; {resp_text}"
+
+    resp_json = json.loads(resp_text)
+
+    if not hasattr(context, "invitation_v2"):
+        context.invitation_v2 = {}
+
+    context.invitation_v2[inviter] = resp_json["invitation"]
+
+@when('"{accepter}" accepts the didcomm v2 invitation from "{inviter}"')
```
nodlesh: The roles here should match what the feature files are stating. Has the role changed to accepter? If so, the feature file needs to change.

nodlesh (changes requested):

aries-test-harness/features/didcomm-v2.feature
Moopli: No, it's not possible to combine things completely - if Acme creates a DIDComm V2 invitation, and Bob accepts that invitation, Acme won't know about Bob until Bob sends Acme a message. The first message Bob sends to Acme isn't part of a standard start-connection protocol (like didexchange is), it can be the start of any protocol that's supported by DIDComm V2 (like issue-credential or present-proof V3). So I needed an additional step for Acme to report its new connection to the test harness, after Acme receives its first message from Bob.

I can certainly combine the create-invitation and accept-invitation steps, though!

Moopli: ^^ Wait I think I misunderstood, if you simply meant create a step that validates that the agents have a DIDComm V2 connection, then yes I can do that!

aries-test-harness/features/steps/didcomm-v2.py
Moopli: fixed :+1:

Moopli:
@nodlesh @TimoGlastra I decided to remove the issue-credential test from my PR, as @DRK3 is preparing a followup PR that adds WACI + DIDComm V2 tests, and he's separating issue credential v3 from v2. I've also moved the endpoints over to oob-v2 instead of didcomm-v2, you raised good points.

aries-backchannels/afgo/afgo_backchannel.py
```diff
@@ -621,6 +625,80 @@ async def handle_agent_POST(self, command: BackchannelCommand):
 
         return (200, "{}")
 
+    async def handle_oob_v2_POST(self, command: BackchannelCommand):
```
TimoGlastra, line 628: Not necessarily for this PR, but I created a way to build new topics in separate files to not make the backchannel file any larger than it already is. E.g. you can see an example for the mediation routes here: https://github.com/hyperledger/aries-agent-test-harness/blob/main/aries-backchannels/afgo/routes/mediation_routes.py

What I like about this approach is that it allows you to search for an endpoint and get the function where it is implemented (can be quite hard sometimes with the xxx_backchannel file). You can find everything related to a single topic in the file, and it gives you a separate handler per endpoint (instead of doing everything in once gigantic method).

I think it greatly reduces the complexity of the code, as we don't have to do all the topic translations etc... which hide what is actually happening.

aries-test-harness/features/oob-v2.feature
```diff
@@ -0,0 +1,21 @@
+@DIDComm-V2 @UsesCustomParameters
```
TimoGlastra:
```suggestion
@DIDCommV2 @OobV2 @UsesCustomParameters
```

aries-test-harness/features/oob-v2.feature
```diff
@@ -0,0 +1,21 @@
...
+Feature: DIDComm V2 Establishing Connections (with Issue Credential example)
+
+   @T001-DIDComm-V2
```
TimoGlastra:
```suggestion
   @T001-OobV2
```

TimoGlastra (dismissed):

aries-test-harness/features/oob-v2.feature
Moopli: done!

aries-test-harness/features/oob-v2.feature
Moopli: done!

aries-test-harness/features/oob-v2.feature
```diff
@@ -0,0 +1,21 @@
+@DIDComm-V2 @OobV2 @UsesCustomParameters
+Feature: DIDComm V2 Establishing Connections (with Issue Credential example)
```
nodlesh: There isn't an Issue Credential example anymore. Can we remove this?

aries-test-harness/features/oob-v2.feature
```diff
@@ -0,0 +1,21 @@
...
+         | Bob  | invitee  |
+
+      And "Acme" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Bob" is running with parameters "{"mime-type":["didcomm/v2"]}"
+
+      When "Acme" creates a didcomm v2 invitation
+      And "Bob" accepts the didcomm v2 invitation from "Acme"
+
+      # Bob must send a message to acme
+      When "Bob" requests mediation from "Acme"
```
nodlesh: `And` not` When` here.

nodlesh (dismissed):
A couple of minor changes in the feature file, but we can merge this without those changes if you want, unless you want to quickly push them in.

aries-test-harness/features/oob-v2.feature
Moopli: done!

aries-test-harness/features/oob-v2.feature
Moopli: done!

nodlesh (approved):
Looks good! 😎👍

PR #445, opened 2022-03-04
enhancement: Starting only agents given in cmd

nodlesh:

This PR modifies manage to only start agents specifically stated in the command. Previously it would attempt to start all roles even if they we not given. Now we can do commands like the following to just start agents needed. 
`./manage start -a acapy-main` Only starts acme as acapy-main. All other roles are not started.
`./manage start -a acapy-main -b afgo-interop` Only starts acme and bob. 
`./manage start -a afgo-interop -d acapy-main` This still works as before, acme is started as afgo, and all other roles are started as acapy-main. 

This enhancement is being done to support aries-mobile-test-harness that can use these agents as services when running mobile tests. It will not need 4 agents running so limiting to only what is asked for is required.

### DIFF
--- a/manage
+++ b/manage
@@ -730,30 +730,34 @@ startHarness(){
   if [[ "$ACME" != "none" ]]; then
     export ACME_AGENT=${ACME_AGENT:-${ACME}-agent-backchannel}
     startAgent Acme acme_agent "$ACME_AGENT" "9020-9029" 9020 9021 "$AIP_CONFIG" "$ACME"
-    waitForAgent Acme 9020
-    echo
   fi
-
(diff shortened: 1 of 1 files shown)


TimoGlastra (commented):
Maybe we can look at starting all needed agents first and then waiting for the agents to start. This is the current behaviour and making the startup of agents serial instead of parallel will probably increase the (already long) startup time of the test harness.

nodlesh:
Yes, that is a good idea @TimoGlastra.

nodlesh:
Thanks @TimoGlastra, I've updated the PR.

TimoGlastra (approved):
Nice!

PR #447, opened 2022-03-10
WACI Issuance tests

DRK3:
Basic WACI issuance tests using DIDComm V2 and Issue Credential V3. Spec for reference: https://identity.foundation/waci-presentation-exchange/#issuance-2

Also added the settings folder used by JetBrains IDEs to the gitignore.

Signed-off-by: Derek Trider <Derek.Trider@securekey.com>

### DIFF
--- a/aries-test-harness/features/data/waci_issue_credential_attachment.json
+++ b/aries-test-harness/features/data/waci_issue_credential_attachment.json
@@ -1,0 +1,66 @@
+[
+  {
+    "id": "df3311ef-894d-4092-9227-22f092d19356",
+    "media_type": "application/json",
+    "lastmod_time": "0001-01-01T00:00:00Z",
+    "data": {
+      "json": {
+        "@context": [
+          "https://www.w3.org/2018/credentials/v1",
+          "https://identity.foundation/credential-manifest/fulfillment/v1"
+        ],
+        "credential_fulfillment": {
+          "id": "a30e3b91-fb77-4d22-95fa-871689c322e2",
+          "manifest_id": "dcc75a16-19f5-4273-84ce-4da69ee2b7fe",
+          "descriptor_map": [
+            {
+              "id": "driver_license_output",
+              "format": "ldp_vc",
+              "path": "$.verifiableCredential[0]"
+            }
+          ]
+        },
+        "proof": {
+          "created": "2021-06-07T20:02:44.730614315Z",
+          "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..NVum9BeYkhzwslZXm2cDOveQB9njlrCRSrdMZgwV3zZfLRXmZQ1AXdKLLmo4ClTYXFX_TWNyB8aFt9cN6sSvCg",
+          "proofPurpose": "authentication",
+          "type": "Ed25519Signature2018",
+          "verificationMethod": "did:orb:EiA3Xmv8A8vUH5lRRZeKakd-cjAxGC2A4aoPDjLysjghow#tMIstfHSzXfBUF7O0m2FiBEfTb93_j_4ron47IXPgEo"
+        },
+        "type": [
+          "VerifiablePresentation",
+          "CredentialFulfillment"
+        ],
+        "verifiableCredential": [
+          {
+            "@context": [
+              "https://www.w3.org/2018/credentials/v1"
+            ],
+            "credentialSubject": {
+              "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+              "license": {
+                "dob": "07/13/80",
+                "number": "34DGE352"
+              }
+            },
+            "id": "https://ontario.ca/claims/DriversLicense",
+            "issuanceDate": "2010-01-01T19:53:24Z",
+            "issuer": "did:foo:123",
+            "proof": {
+              "created": "2021-06-07T20:02:44.730614315Z",
+              "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..NVum9BeYkhzwslZXm2cDOveQB9njlrCRSrdMZgwV3zZfLRXmZQ1AXdKLLmo4ClTYXFX_TWNyB8aFt9cN6sSvCg",
+              "proofPurpose": "assertionMethod",
+              "type": "Ed25519Signature2018",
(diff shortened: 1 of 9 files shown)


DRK3:
This is my first PR for the Aries test harness, so feedback will be very welcomed! :)

DRK3:
cc @Moopli

TimoGlastra (changes requested):
This is a great addition @DRK3! I have a few questions on how WACI / DIDComm / Issue Credential V3 work.

This PR adds completely new endpoints for the WACI flow, however as I understand it WACI uses V3 of the Issue Credential protocol with a custom attachment format (`dif/credential-manifest`). For the issue credential v2 protocol we also use custom attachments (indy and jsonld), so I'm not sure if doing this different for WACI is the right way to go. Could we generalize the API endpoints and approach it the same way as we did with issue-credential v2?

I see in the `issueCredentialWRecIDOperationTranslationDict` dict that you map the custom waci endpoints to just the normal `accept-offer` etc... endpoints in AFGO, leading me to believe you're also not using custom WACI endpoints in AFGO. I'm good with adding `@WACI` tags to the tests so it becomes easier to test for WACI compliance, but think we can solve it without creating custom waci endpoints. What are your thoughts on this?

I'm a bit confused about WACI and DIDComm in general. It seems the new issue credential and present proof v3 protocols are defined as part of WACI. Shouldn't they live under the new didcomm.org repository?

swcurran:
Looks good in running the test -- nice!  

I agree with @TimoGlastra that we should look at not making WACI a separate thing from the rest of the suite, but in line with the versioning we have built into the Test Harness.  WACI is (AFAIK):

- V3 versions of the V2 Issue Credential and Present Proof protocols (RFC0453, RFC0454) changed only by the requirements of DIDComm v2. However, they are not (AFAIK) formalized anywhere. I personally would prefer to see them as new Aries RFCs, but others might not be happy with that.
- The use of DIDComm V2 vs. DIDComm V1, which implies an evolution of the OOB/DIDExchange protocols (RFC0434, RFC0025)
- New attachments on V3 of the Issue Credential protocol to handle Credential Manifest.

Interested in your thoughts on that @DRK3 .

DRK3:
Hi @TimoGlastra and @swcurran!

I agree with your comments about WACI not being a separate endpoint - I'll work on that and update my PR accordingly.

As for this question:

> I'm a bit confused about WACI and DIDComm in general. It seems the new issue credential and present proof v3 protocols are defined as part of WACI. Shouldn't they live under the new didcomm.org repository?

I didn't really work on the spec myself, but I think my colleague @rolsonquadras may be able to help answer that, as he worked directly on the spec.

DRK3:
With regards to 

> V3 versions of the V2 Issue Credential and Present Proof protocols (RFC0453, RFC0454) changed only by the requirements of DIDComm v2. However, they are not (AFAIK) formalized anywhere. I personally would prefer to see them as new Aries RFCs, but others might not be happy with that.
The use of DIDComm V2 vs. DIDComm V1, which implies an evolution of the OOB/DIDExchange protocols (RFC0434, RFC0025)
New attachments on V3 of the Issue Credential protocol to handle Credential Manifest.

Again, perhaps @rolsonquadras can weigh in here.

rolsonquadras:
@swcurran 
> * V3 versions of the V2 Issue Credential and Present Proof protocols (RFC0453, RFC0454) changed only by the requirements of DIDComm v2. However, they are not (AFAIK) formalized anywhere. I personally would prefer to see them as new Aries RFCs, but others might not be happy with that.

Yes - the difference is in the message structure (DIDComm v1 vs DIDComm v2). Currently, the issue_credential v3 and present_proof v3 live in DIF WACI repo https://github.com/decentralized-identity/waci-presentation-exchange. 

I think, there was a question about whether it should be in Aries RFCs or DIF DIDComm group. Tagging @TelegramSam @OR13 @brentzundel @bumblefudge @troyronda to see if they have any thoughts on this.

OR13 (dismissed):
Thanks for this PR.

aries-test-harness/features/data/waci_issue_credential_attachment.json
```diff
@@ -0,0 +1,63 @@
...
+            "issuanceDate": "2010-01-01T19:53:24Z",
+            "issuer": "did:foo:123",
+            "proof": {
+              "created": "2021-06-07T20:02:44.730614315Z",
+              "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..NVum9BeYkhzwslZXm2cDOveQB9njlrCRSrdMZgwV3zZfLRXmZQ1AXdKLLmo4ClTYXFX_TWNyB8aFt9cN6sSvCg",
+              "proofPurpose": "assertionMethod",
+              "type": "Ed25519Signature2018",
+              "verificationMethod": "did:orb:EiA3Xmv8A8vUH5lRRZeKakd-cjAxGC2A4aoPDjLysjghow#tMIstfHSzXfBUF7O0m2FiBEfTb93_j_4ron47IXPgEo"
+            },
+            "type": "VerifiableCredential"
```
Exulansis: I believe this property must be an array, e.g. `["VerifiableCredential", "EUDriversLicense"]`, but perhaps I am mistaken.

As mentioned in the [VC specification](https://www.w3.org/TR/vc-data-model/#types):
> All [credentials](https://www.w3.org/TR/vc-data-model/#dfn-credential), [presentations](https://www.w3.org/TR/vc-data-model/#dfn-presentations), and encapsulated objects MUST specify, or be associated with, additional more narrow [types](https://www.w3.org/TR/vc-data-model/#dfn-type) (like UniversityDegreeCredential, for example) ...

DRK3: You're right - I'll push an update to my PR soon!

DRK3:
Hi @TimoGlastra @swcurran @Exulansis,

I've updated my PR based on your feedback and also rebased with the latest changes on main. Can you take another look and let me know if you have any other feedback? Thanks!

aries-test-harness/features/data/waci_issue_credential_attachment.json
Exulansis: I believe the correct value would be `["VerifiableCredential", "EUDriversLicense"]`. I.e. the spec requires the first value to be the more generic `VerifiableCredential` type, and the follow-up values more concrete sub-types.

Perhaps this usage is also okay, not sure if this would cause issues with other existing VC / Agent implementations.

DRK3: My bad - you're absolutely correct. I just updated the PR with the correct `type` - let me know if it looks good.

Exulansis: Looks good! Thank you.

I think this conversation can be marked as resolved.

TimoGlastra (approved):
LGTM, nice!

PR #449, opened 2022-03-16
feat: present-proof v3 (WACI share) test and afgo support

Moopli:

### DIFF
--- a/aries-backchannels/afgo/afgo_backchannel.py
+++ b/aries-backchannels/afgo/afgo_backchannel.py
@@ -174,6 +174,7 @@ class AfGoAgentBackchannel(AgentBackchannel):
         self.proofOperationTranslationDict = {
             "prepare-json-ld": "",
             "send-proposal": "send-propose-presentation",
+            "accept-proposal": "accept-propose-presentation",
             "send-request": "send-request-presentation",
             "send-presentation": "accept-request-presentation",
             "verify-presentation": "accept-presentation",
@@ -190,6 +191,7 @@ class AfGoAgentBackchannel(AgentBackchannel):
             "issue-credential-v2": "/issuecredential/",
             "proof": "/presentproof/",
             "proof-v2": "/presentproof/",
+            "proof-v3": "/presentproof/",
         }
 
         # Generic AATH parameter to AFGO commandline parameter
@@ -364,6 +366,8 @@ class AfGoAgentBackchannel(AgentBackchannel):
         if "piid" in message["message"]["Properties"]:
             thread_id = message["message"]["Properties"]["piid"]
             push_resource(thread_id, "present-proof-states-msg", message)
+
+            await push_message_stack("present-proof-states-msg" + "," + thread_id, message)
         else:
             raise Exception(
                 f"piid not found in present_proof_states Webhook Message: {json.dumps(message)}"
@@ -556,6 +560,10 @@ class AfGoAgentBackchannel(AgentBackchannel):
 
         elif command.topic == "proof" or command.topic == "proof-v2":
             (resp_status, resp_text) = await self.handle_present_proof_POST(command)
+            return (resp_status, resp_text)
+
+        elif command.topic == "proof-v3":
+            (resp_status, resp_text) = await self.handle_present_proof_v3_POST(command)
             return (resp_status, resp_text)
 
         # Handle out of band POST operations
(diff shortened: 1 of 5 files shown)


aries-backchannels/afgo/afgo_backchannel.py
```diff
@@ -174,6 +174,7 @@ def __init__(
         self.proofOperationTranslationDict = {
             "prepare-json-ld": "",
             "send-proposal": "send-propose-presentation",
+            "accept-proposal": "accept-propose-presentation",
             "send-request": "send-request-presentation",
```
TimoGlastra, line 178: What's the difference between accept-proposal and send-request?

TimoGlastra (commented):
Nice @Moopli. Could you possibly add API definitions for the new endpoints? This will help future implementors :)

aries-test-harness/features/steps/present-proof-v3.py
```diff
@@ -0,0 +1,317 @@
...
+from agent_test_utils import (
+    get_relative_timestamp_to_epoch,
+    amend_presentation_definition_with_runtime_data,
+)
+from distutils.util import strtobool
+
+
+@given(
+    '"{issuer}" has issued a credential with {credential_data} to "{prover}" using issue-credential v3'
+)
```
nodlesh: This code in this step implementation is very close to what is already in 0454-present-proof-v2.py. Could we not modify that original step to handle both?

aries-test-harness/features/present-proof-v3.feature
```diff
@@ -0,0 +1,33 @@
+@PresentProofV3 @DIDComm-V2 @UsesCustomParameters
+Feature: RFC 0454 Aries agent present proof v3
```
nodlesh: Does it make sense to reference RFC 0454 here? Do we not need a different reference of either another RFC or some specification? Does RFC 0510 even make sense here or as a tag in the scenarios?

aries-test-harness/features/present-proof-v3.feature
```diff
@@ -0,0 +1,33 @@
...
+   @T001-PresentProofV3 @RFC0510 @WACI @CredFormat_JSON-LD @Schema_Citizenship_Context @CredProposalStart
+   Scenario Outline: Present Proof of specific types and proof is acknowledged with a Citizenship credential type with a DID Exchange Connection
+      Given "2" agents
+         | name  | role     |
+         | Faber | verifier |
+         | Bob   | prover   |
+      And "Acme" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Bob" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Faber" is running with parameters "{"mime-type":["didcomm/v2"]}"
+      And "Acme" has issued a credential with <credential_data> to "Bob" using issue-credential v3
```
nodlesh: I think `issuer credential v3` is not adding anything to the step declaration here. We already know it will be V3 based on the tags, most importantly the `@DIDcomm-V2` tag. We could capture that tag in code and use the proper issue_cred v3. The same pattern we did for Issue Credential v1 and v2 should be followed IMO. Thoughts?

aries-test-harness/features/present-proof-v3.feature
```diff
@@ -0,0 +1,33 @@
...
+
+      When "Faber" creates a didcomm v2 invitation with goal code "streamlined-vp"
+      And "Bob" accepts the didcomm v2 invitation from "Faber"
+      And "Bob" sends a v3 presentation proposal to "Faber"
```
nodlesh: Same issue here with v3 Presentation as my issue with Issue Credential v3 above.

nodlesh (changes requested):
Looks good. Just a few questions.

aries-backchannels/afgo/afgo_backchannel.py
Moopli: Present-proof v3, like v2, can start with a request message or a proposal - `send-request` starts a new thread, starting the protocol with the request, while `accept-proposal` continues the thread started by a proposal.

aries-test-harness/features/present-proof-v3.feature
Moopli: I should remove the mention of 0454, yes, but 0510 is indeed a valid reference - the WACI spec mentions RFC 510 for the definition of the presentation-exchange format.

aries-test-harness/features/present-proof-v3.feature
Moopli: Don't we have different steps for issue-credential v1 and v2 already? I was continuing the pattern of having separate steps for separate protocols. I do agree though that it would be cleaner for the features if these were combined, and if the step implementation used a tag to determine behaviour.

nodlesh: That is a valid point, but they were separated because in v2 we went on to implement the input data into the step. Issue Credential V1 tests don't have that. I can image if we went back to refactor those tests to include the data tables and examples to pass data in, we would probably reuse the code in issue credential v2.

aries-backchannels/afgo/afgo_backchannel.py
TimoGlastra: Hmm I thought they were combined in one endpoint. E.g. if you look at send offer is issue credential v2 it allows two ways of calling it: https://aries-interop.info/api#operation/IssueCredentialV2SendOffer. One with an id (the thread id) and one with data

Moopli: Just addressed this, can you review?

aries-test-harness/features/steps/present-proof-v3.py
Moopli: Done!

aries-test-harness/features/present-proof-v3.feature
Moopli: fixed this!

aries-test-harness/features/present-proof-v3.feature
Moopli: Just finished merging the v2/v3 steps, using tag logic as you suggested :)

aries-test-harness/features/present-proof-v3.feature
Moopli: Done!

swcurran:
Nice work!  Ran through it and looks good.  I'll leave it to @nodlesh and @TimoGlastra to complete their review.

TimoGlastra (approved):
LGTM

nodlesh (approved):

OR13:
We've been tracking this work here as well: https://github.com/w3c-ccg/traceability-interop/issues/17

Seems we are almost ready to have a conversation about implementing support for WACI-PEx DIDComm in the VC API.

PR #451, opened 2022-03-18
fix: tag rfc0025+rfc0160 test with @RFC0160 so afgo suite skips it

Moopli:


Moopli:
@TimoGlastra #448 broke afgo's suite, by adding a second set of tests that use RFC0160 but without the RFC0160 tag, this just adds the missing tag so afgo's suite correctly skips them.

TimoGlastra (approved):
Whoops, sorry about that!

Issue #457, opened 2022-03-28
Proposal to fix intermittent fails of tests: Add action to run all runset actions and remove cron triggers


swcurran:
We continue to get intermittent test runs being cancelled.  Since the same tests pass one day, cancelled the next and run to completion the day after, it seems likely the problem is one of resource usage -- we have too many tests running in parallel, despite attempts to spread out the tests.

I'm wondering if someone with a depth in GitHub experience could create a job that triggers sequential runs of multiple other jobs to see if that is a viable solution.  If that works we would have each of the runsets Actions (e.g. [test-harness-acapy-afgo.yml](https://github.com/hyperledger/aries-agent-test-harness/blob/main/.github/workflows/test-harness-acapy-afgo.yml)) changed to remove the cron (e.g. manually triggered only).  Then we would have one job that sequentually triggers all of the Actions, plus the "generate results" action.  It means having to keep the names of the actions both as files and listed in the "Run Everything" Action, but that would be easier to manage then the current cron settings on each job.

@WadeBarnes -- if you could try starting on this -- e.g. investigate how to trigger Actions in sequence from an Action -- I can do the rest of the updates to put everything in place (which admittedly is pretty easy :-) ). 

FYI @nodlesh 

Thanks!

WadeBarnes:
I'll have a look.

WadeBarnes:
The initial attempts failed, I ran into a limitation on the number of reusable workflows you can call from your main workflow, and the calling depth when trying to work around that limitation.  Though I have another idea.

WadeBarnes:
Forcing the jobs to run sequentially does not work around the workflow reference limit either.

swcurran:
Could you share what you are trying?  I found a couple of things that suggested approaches, but my understanding of GHAs is not good enough?

WadeBarnes:
See the runs here; https://github.com/WadeBarnes/aries-agent-test-harness/actions

WadeBarnes:
Current iteration looks like this.; https://github.com/WadeBarnes/aries-agent-test-harness/blob/feature/test-control/.github/workflows/run-tests.yml.  I was looking for something that could be done fast with minimal changes to prove out the resource theory.

WadeBarnes:
There is a slightly more complicated from of the same thing I can try that might work.

swcurran:
[...]

PR #459, opened 2022-03-30
Refactor test harness flow control

WadeBarnes:
- Switch to a main test workflow to control the flow and execution of the test harnesses.
- Uses a throttled matrix strategy to control the number of concurrent test harness workflows running at one time.
- Supports:
  - dynamic discovery of test harness workflows.
  - exclusion of specific test harness workflows.

### DIFF
--- a/.github/workflows/generate-summary-results.yml
+++ b/.github/workflows/generate-summary-results.yml
@@ -1,15 +1,12 @@
 name: generate-aath-summary-results
 on:
   workflow_dispatch:
-  schedule:
-    - cron: "45 3 * * *"
 defaults:
   run:
     shell: bash
 jobs:
   test:
     runs-on: ubuntu-latest
-    if: (github.event_name == 'schedule' && github.repository == 'hyperledger/aries-agent-test-harness') || (github.event_name != 'schedule')
     steps:
       - name: checkout-test-harness
         uses: actions/checkout@v2
--- a/.github/workflows/test-harness-acapy-afgo.yml
+++ b/.github/workflows/test-harness-acapy-afgo.yml
@@ -18,15 +18,12 @@ name: test-harness-acapy-afgo
 # End
 on:
   workflow_dispatch:
-  schedule:
-    - cron: "5 0 * * *"
 defaults:
   run:
     shell: bash
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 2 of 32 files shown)


WadeBarnes:
This PR is based on the investigation and testing preformed for this issue; https://github.com/hyperledger/aries-agent-test-harness/issues/457

Please review and provide feed back.

Items to note:
- ~max-parallel: 7~
  - ~I have it running 7 test harnesses concurrently ATM for testing only.  We may want to bring that down to 5 and see how things go from there.  I had it set to 5 on previous runs and it took a few hours to run through all the test harnesses.~
- ~I need to remove the `push` event from the `test-harness-runner` workflow (test-runner.yml) before merging.~
- ~If we proceed with this approach (which I think is at least better than the existing approach) we'll need Ry to provision a public repo scoped PAT (named REPO_ACCESS_TOKEN) for use with the new workflow before merging.~
- All the above has been updated.

WadeBarnes:
Latest test run can be found here; https://github.com/WadeBarnes/aries-agent-test-harness/actions/runs/2065794795
- Note: - All of the tests fail on the report upload step because I don't have Allure credentials setup so I don't pollute the reports.

swcurran:
Way cool! I think that is excellent.

Only changes I would suggest is removing some of the comments in the script itself, and perhaps adding a section to the README (or reference to another document) about adding new runsets -- GHAs that run tests.

Love to be able to run it tonight.

WadeBarnes:
> Only changes I would suggest is removing some of the comments in the script itself, and perhaps adding a section to the README (or reference to another document) about adding new runsets -- GHAs that run tests.

@swcurran, I've updated the workflow and added some documentation.

- max-parallel has also been set to 5 and the `push` event has been removed.

WadeBarnes:
@ryjones, would you be able to add a public repo scoped PAT named **REPO_ACCESS_TOKEN** to this repository for use with this new test harness workflow?  The PAT only needs the `public_repo` permission, provided the account it's from already has access to the repo.

ryjones:
@WadeBarnes done

WadeBarnes:
> @WadeBarnes done

Thanks @ryjones

WadeBarnes:
@swcurran, @nodlesh, @ianco, This one is ready to go if you're happy with it.  The updates and the necessary secret are all in place.

swcurran (approved):
Brilliant!!  Great work!

swcurran:
Made it in time for today's run, I think :-).  Let's see how it goes.

PR #463, opened 2022-03-31
Fix repository references

WadeBarnes:
- The GitHub data for scheduled workflows does not contain the github.event.repository attributes.  Therefore the repository name needs to be parsed from the repository attribute.


WadeBarnes:
@swcurran, This is the fix for the error seen in the scheduled run.

swcurran (approved):

PR #465, opened 2022-04-01
feat: support did:sov and did:key

TimoGlastra:
Adds support for RFC 0453 and RFC 0454 tests using did:sov and did:key between AFGO agents. did:key was already supported, but AFGO was actually using did:orb under the hood when issuing. Now AFGO also supports issuing with did:key dids. Also added support in ACA-Py for issuing JSON-LD creds using did:sov dids.

This gives us the following support matrix:

|        | AFGO                            | ACA-Py                          |
|--------|---------------------------------|---------------------------------|
| Issue  | `did:orb`, `did:key`            | `did:sov`, `did:key`            |
| Verify | `did:orb`, `did:sov`, `did:key` | `did:orb`, `did:sov`, `did:key` |


swcurran (approved):

swcurran:
LGTM - nice work!

PR #466, opened 2022-04-04
Adjust parallelism of tests

WadeBarnes:
- Bump it down one to see if there is an affect on the cancellation of test-harness-acapy.yml (test-harness-acapy-acapy).


nodlesh (approved):

swcurran:
Worth a try for sure -- let's see how it goes.

I suspect the issue is something specific with this run and test.  Not absolutely clear, since we have had different runsets fail over time. But recently this specific failure (when it happens) has been consistent as to the place.

PR #467, opened 2022-04-06
feat: support revocation notification v1

TimoGlastra:
Adds support for revocation notification v1 in the revocation tests. For ACA-Py it currently uses an incorrect value for the `thread_id` in the revocation notification, so this is also being used in the AATH tests. I don't think there's an implementation that implemented the actual RFC. It also supports the correct thread_id when checking if a revocation notification has been received, so if other agents do support the RFC variant it will also work (but not with ACA-Py as ACA-Py doesn't follow the RFC).


regiseloi-idlab (commented):
This looks good, thank you for explaining how to handle the inconsistencies between the way revocation notification v1 is defined in the RFC and how it's been implemented so far in ACA-Py (comment on [aries-test-harness/features/steps/0011-0183-revocation.py](https://github.com/hyperledger/aries-agent-test-harness/pull/467/files#diff-adce4e363b8cba7b891776d3fad6420eaabaa96aad04e92f50d27281b6a870ae)).

swcurran (approved):
LGTM

PR #443, opened 2022-03-02
Invalid link to Advanced Test Reporting

rqpb:
This pull request fixes an invalid link to Advanced Test Reporting in README.md.

PR #470, opened 2022-04-08
Fix bad link as highlighted in #443

swcurran:


swcurran:
Replaces #443 as that change was not DCO signed.  Thanks for noting the error @rqpb!

nodlesh (approved):

PR #474, opened 2022-04-10
feat: oob connectionless for issue cred and present proof

### LINKED ISSUE #90: Finish Connectionless Proof Test Scenario(s) to AATH
This was attempted in Aug 2020. There is code in the AATH to make this work. In submission related to this issue. It was determined that Acapy at the time did not support the connectionless scenario is played as the prover. This work has been put on hold until Acapy has this fully implemented. 

The main Story for this work was #11.

TimoGlastra:
Adds connectionless support for both issue credential v1/v2 and present proof v1/v2 using the oob protocol. Also removed the old AIP 1 connectionless code as it was never used.

Draft for now as it's built against my fork of ACA-Py until it get's merged into main.

This only handles attachments in OOB invitations for connectionless. Probably good to also add tests for receiving oob invitations with creating a connection and processing an attachments (Should be quite easy now with all substeps being defined).

One thing I ran into and we should probably change is that AATH is build on the expectation that the oob receive invitation will return a connection record and also connection state instead of oob state. Probably because that's how it was implemented in ACA-Py. For now I've tricked the ACA-Py backchannel into returning the state it expects for now (mapping `prepare-response` to `invitation-received`), but would be good to change this to be more in line with the RFC (having less ACA-Py driven backchannel API design would be good in general I think)

We may want to move some tests from the DIDExchange tests to the oob tests now that there is a dedicated file for out of band


nodlesh (dismissed):
Thanks for updating these!

nodlesh:
Sorry, didn't notice it's draft.

ianco:
Noticed a few conflicts, related to code formatting so maybe a different black version?

TimoGlastra:
I think we're not enforcing black for this repo, so that's probably why? I'll see to address the conflicts tomorrow!

swcurran:
Let us know when it is past draft. A little worried that existing tests broke because of the ACA-Py PR. Does that mean a breaking change for controllers or is it just AATH?

TimoGlastra:
Away from my computer until Tuesday, will take a look at the conflicts then

swcurran:
Adding note on this PR for completeness.  The [ACA-Py "connectionless" PR](https://github.com/hyperledger/aries-cloudagent-python/pull/1710) has been merged, and that generated a number of ACA-Py to ACA-Py DID Exchange errors.  Testing on this branch does not seem to fix the errors.

aries-backchannels/acapy/requirements-main.txt
```diff
@@ -1,2 +1,2 @@
-aries-cloudagent[indy, bbs, askar]@git+https://github.com/hyperledger/aries-cloudagent-python@main
+aries-cloudagent[indy, bbs, askar]@git+https://github.com/TimoGlastra/aries-cloudagent-python@fix/oob-connectionless
```
swcurran: This needs to be changed before merging.

swcurran (changes requested):

swcurran (approved):
Tested for all ACA-Py - ACA-Py tests.  Working.  w00t!!

PR #475, opened 2022-04-10
fix: waci tags and did method

TimoGlastra:
The waci tests weren't specifying any did method, causing the flow to fail after my recent pr to extend did methods


DRK3 (approved):

nodlesh (approved):
👍

PR #486, opened 2022-05-06
Add container logs upload step to common GitHub test action.

lauravuo-techlab:
It would be beneficial sometimes to inspect the logs for the different containers that were run during the test flow in GitHub Actions. This change will collect and archive the container logs as artifacts.

Signed-off-by: Laura Vuorenoja <laura.vuorenoja@op.fi>

### DIFF
--- a/actions/run-test-harness-wo-reports/action.yml
+++ b/actions/run-test-harness-wo-reports/action.yml
@@ -37,12 +37,28 @@ runs:
       run: ./manage build ${{ inputs.BUILD_AGENTS }}
       shell: bash
       working-directory: test-harness
+    - name: start-harness-agents
+      run: ./manage start ${{ inputs.TEST_AGENTS }}
+      shell: bash
+      working-directory: test-harness
     - name: run-test-harness-acapy
-      run: PROJECT_ID=${{inputs.REPORT_PROJECT}} ./manage run ${{ inputs.TEST_AGENTS }} ${{ inputs.REPORTING }} ${{ inputs.OTHER_PARAMS }} ${{ inputs.TEST_SCOPE }}
+      run: PROJECT_ID=${{inputs.REPORT_PROJECT}} ./manage test ${{ inputs.TEST_AGENTS }} ${{ inputs.REPORTING }} ${{ inputs.OTHER_PARAMS }} ${{ inputs.TEST_SCOPE }}
       shell: bash
       env:
         NO_TTY: "1"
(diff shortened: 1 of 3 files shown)


actions/run-test-harness-wo-reports/action.yml
```diff
@@ -37,12 +37,28 @@ runs:
       run: ./manage build ${{ inputs.BUILD_AGENTS }}
       shell: bash
       working-directory: test-harness
+    - name: start-harness-agents
+      run: ./manage start ${{ inputs.TEST_AGENTS }}
+      shell: bash
+      working-directory: test-harness
     - name: run-test-harness-acapy
-      run: PROJECT_ID=${{inputs.REPORT_PROJECT}} ./manage run ${{ inputs.TEST_AGENTS }} ${{ inputs.REPORTING }} ${{ inputs.OTHER_PARAMS }} ${{ inputs.TEST_SCOPE }}
+      run: PROJECT_ID=${{inputs.REPORT_PROJECT}} ./manage test ${{ inputs.TEST_AGENTS }} ${{ inputs.REPORTING }} ${{ inputs.OTHER_PARAMS }} ${{ inputs.TEST_SCOPE }}
```
nodlesh, line 45: If you are using the `test` command then I don't think you need the `inputs.TEST_AGENTS`

lauravuo-techlab: Apparently this param is also needed. Otherwise the test will fail and output following:


`All agents Acme, Bob, Faber and Mallory must be set to one of: acapy acapy-main afgo-interop afgo-master aries-vcx dotnet dotnet-master findy javascript mobile verity .`

nodlesh (approved):
Looks great! Thanks.

actions/run-test-harness-wo-reports/action.yml
nodlesh: @lauravuo-techlab Was the change to `manage test` from `manage run` needed to be able to collect and archive the docker logs? We are needing to revert back to `manage run` for proper environment.properties generation and interop summary processing. Will this be ok for handling the logs?

lauravuo-techlab: Ok! Sorry about the hassle, I was not aware of the environment properties generation. Yes, the automated docker logs collection works only when the containers are running and as `run` shuts down the agent containers, they are not stored if `run` is used.

PR #487, opened 2022-05-07
fix(afj): issue credential controller

TimoGlastra:
Fixes the afj backchannel with breaking changes introduced by merging icv2 support.

Support for v2 will be added in a later PR (there are some issues currently that make it not work)


nodlesh (approved):
LGTM

PR #490, opened 2022-05-13
Add special case handling when there are no Allure reports for a runset and add debug lines to site generator script

swcurran:


swcurran:
Fixed handling when there are no results available.

nodlesh (approved):

PR #491, opened 2022-05-13
Hack to make dotnet test agent work by making it the same as dotnet-master

swcurran:

### DIFF
--- a/aries-backchannels/dotnet/Dockerfile.dotnet
+++ b/aries-backchannels/dotnet/Dockerfile.dotnet
@@ -1,16 +1,49 @@
 # Final image needs indy
+# FROM streetcred/dotnet-indy:1.15.0 AS base
+# WORKDIR /app
+
+# FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build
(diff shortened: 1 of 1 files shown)


swcurran:
The recent fix to make the "dotnet-master" test agent (built from main branch) to work broke the "dotnet" test agent.  The fix changed the structure of the .NET project and until a new version is published that reflects the new structure, the "dotnet" agent won't work. There are several ways to fix this --- publish a new .NET version (not our call), make a separate backchannel for the two versions, ignore/delete the "dotnet" test agent or put a compensation in.

I went with just making the two test agents the same -- but keeping "dotnet" test agent by making the two Dockerfiles the same.  So both "dotnet" and "dotnet-master" are identical agents running the latest dotnet code in the repo.

Later we can revert the change, once the repo and published versions use the same structure.

swcurran:
@nodlesh -- can you push this through?  I'd like to run the dotnet tests today, if possible.

nodlesh (approved):

nodlesh:
I think you need to update your branch before merge?

PR #494, opened 2022-05-13
Rename verity workflows to skip them so the summary results are not generated

swcurran:
When workflows need to be skipped, as with verity, the skipping has to happen in both the test-runner.yml script and in the test results generator. Renaming the files seems a better way to do that -- and it's more obvious what is happening.


swcurran:
Note -- the reference in the test-runner is still there, but it no longer does anything.

swcurran:
@WadeBarnes -- do you like this approach vs. the pattern matching you have in your script.

I also have an alternative way to do it by just adding a "SKIP" line in the action.  That was added to trigger a skip in the processing and would work.  However, you have to remember to both update the "SKIP" line (for the results generator) and update the "exclude" pattern in the GHA to bypass the script.  I kind of like this renaming of the GHA to make it clear we're skipping it.

WadeBarnes:
@swcurran, I agree renaming the files makes it more obvious.  You should update the `test-runner.yml` to remove the ignore pattern and replace it with a note about renaming the files to be skipped.

swcurran:
@WadeBarnes -- how about that?

WadeBarnes (approved):

PR #495, opened 2022-05-16
Reverted to manage run instead of manage test

nodlesh:

This PR reverts the 'manage test' in the run test action to 'manage run'. This is required to properly generate the environment.properties for reporting and interoperability summary processing. We could add this to the `test` command, but it wasn't really the intent of that command to run on nightly builds - There could be other side effects. 

@lauravuo-techlab if this has an effect on the agent log archiving, then we will have to think of another solution.

### DIFF
--- a/actions/run-test-harness-wo-reports/action.yml
+++ b/actions/run-test-harness-wo-reports/action.yml
@@ -37,12 +37,8 @@ runs:
       run: ./manage build ${{ inputs.BUILD_AGENTS }}
       shell: bash
       working-directory: test-harness
-    - name: start-harness-agents
-      run: ./manage start ${{ inputs.TEST_AGENTS }}
-      shell: bash
-      working-directory: test-harness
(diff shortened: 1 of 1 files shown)


swcurran (approved):
It's not clear that this will make a difference, but let's give it a try.  The "start" ahead of the "test" command is equivalent to the "run", so in theory there is no difference other than the lack of a "stop" command after the "test". However, the data is not getting correctly to Allure and it's not clear why not.

lauravuo-techlab:
@nodlesh The agent logs will not be stored when this change is applied since the docker container logs collection works only for running containers. Maybe we need to reconsider the approach for "manual" logs export (as was done [here](https://github.com/hyperledger/aries-agent-test-harness/pull/416/files)). Although I would prefer using the automated action since it will collect logs for all running containers - and no changes would be needed to log collection if containers are added or removed.

PR #496, opened 2022-05-17
Sort test results

ryjones:
Sort the webpage so the intro is first

Signed-off-by: Ry Jones <ry@linux.com>

### DIFF
--- a/docs/acapy.md
+++ b/docs/acapy.md
@@ -1,3 +1,6 @@
+---
+sort: 2 # follow a certain sequence of letters or numbers
+---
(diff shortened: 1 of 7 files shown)


swcurran:
@ryjones -- the files you are looking at are generated via a GHA, not manually maintained.  If you want the data sorted (and what is it you want sorted?) you need to do that in the `gen-interop.sh`.  Probably easier is to tell me what you want done and I can adjust the script.  I think it is just a matter of putting a hardcoded array in order.

swcurran:
I assume the right thing is to close this and make this change another way.

Issue #499, opened 2022-05-18
dotnet RFC 0037 Present Proof tests getting error - not sending ACK


swcurran:
When I run the "dotnet" as Bob (Holder/Prover) tests for [RFC 0037 Present Proof](https://github.com/hyperledger/aries-rfcs/tree/main/features/0037-present-proof), with both Bob and ACA-Py as the Faber (verifier), the tests (e.g. `@T001.5-RFC0037`) fail with the following error:

` From http://0.0.0.0:9030 Expected state ['done', 'N/A'] but received presentation-sent , with a response status of 200`

When I run the same tests with "acapy-main" as Bob, and either ACA-Py or Aries Framework .NET as the Verifier, the tests work.

I'm pretty sure the reason for this is that the verifiers (ACA-Py and Aries Framework .NET) do not send back to the Holder/Prover an "Ack" to indicate that the proof was received and verified, and the Aries Framework .NET holder/prover stays in the `presentation-sent` state, while the ACA-Py holder/prover transitions to the "done" state immediately after sending the presentation. If that is the case, which is "right" according to the RFC, and hence, what should the tests do?

Here are a series of notes/thoughts/questions on this. I'd like to get feedback from those that know AATH (@nodlesh , @TimoGlastra, @ianco  in particular) to get a view on this.

* Reading the RFC, the drawing has the Verifier sending the ACK, but the text does not say that.
    * Amongst implementers of RFC 0037, is there any question if the ACK is required or not?
* If I assume that an ACK (or PROBLEM-REPORT) is NOT required, I can "fix" the tests by changing the "expected_state" in the test from `"done"` to `["done", "presentation-sent"]` [here](https://github.com/hyperledger/aries-agent-test-harness/blob/e1a6c9569236c2bd4aaf32b296daf71c3b2db53e/aries-test-harness/features/steps/0037-present-proof.py#L346), the tests work, as the holder/prover is in one of the expected states used by the different frameworks.
    * staying in the "presentation-sent" state is a problem for a real holder/prover as they then need a way to clean up this now hanging protocol state object that will never automagically transition to the "done" state.
    * Is this "expect one of several states" a good way to "fix" this issue so that the tests pass? Or does that hide interoperability problems?
* If sending the ACK is required, I suspect that the issue with the tests failing is actually in the AATH .NET and ACA-Py Backchannels -- e.g. the verifier (Faber) is not telling its Aries framework instance to send the ACK message.
    * Is this where the problem exists and how we should address it?

PR #500, opened 2022-05-19
Add flexibility in what is the end-state for Holders for RFC 0037

swcurran:

### DIFF
--- a/aries-test-harness/features/steps/0037-present-proof.py
+++ b/aries-test-harness/features/steps/0037-present-proof.py
@@ -339,11 +339,14 @@ def step_impl(context, verifier):
 @then('"{prover}" has the proof verified')
 def step_impl(context, prover):
     # check the state of the presentation from the prover's perspective
+    # All flexibility in the expected state to include "presentation-sent"
+    #   as Aries frameworks do not send an Ack from the verifier after verification and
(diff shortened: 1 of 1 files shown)


swcurran:
Addresses issue #499 by accepting that the current RFC 0037 implementations do not always have the verifier send an ACK after verification, and as a result, some holders (specifically, Aries Framework .NET) leaves the protocol state at "presentation-sent".  I think this is OK since the protocol is complete.  It is not ideal for the Holder instance, as the Framework or controller would have to figure out how/when to clean up the protocol state object, but that does not represent an interoperability concern.

Note that it could mean that if the verifier does send an ACK and there is a transition by the Holder to the "done" state, it happens after the Test Harness check that is satisfied by a check of that state that returns "presentation-sent", but I don't see a big issue with that.

This will likely fix all of the "dotnet as a holder" tests across all test runsets.

nodlesh (approved):
I think I'm good with this. In the same step we are asserting the presentation verification status anyway, which should be `true`, no matter what the state is at this point.

Could be an enhancement in the future when we move to asking the Frameworks  for supported Capabilities, what their end states would be on each protocol, or instead of hardcoding the RFC states, we ask for a protocol state transition scheme to validate against. Just thinking out loud here.

PR #501, opened 2022-05-26
fix(afj): connection controller after oob

### LINKED ISSUE #497: Test agent "javascript" not starting
Noticed that there are no ACA-Py - AFJ test results being produced.  Ran the script locally and discovered that the JavaScript agent is not starting.  The log is in the next comment. Because of other issues with the test suite, I'm not able to pinpoint when the test agent stopped working, but given the log and the reference to Credential Exchange V1, perhaps it was with the addition of the V2 Credential exchange protocols.

@TimoGlastra -- could you please take a look at this and let us know what is up?

TimoGlastra:
This updates the AFJ backchannel to resolve the issues from the breaking changes introduced after merging oob/didexchange support. I haven't had time to fully check if everything works, but this should at least fix the issue of AFJ not building at all.

Fixes #497

### DIFF
--- a/aries-backchannels/javascript/server/src/controllers/ConnectionController.ts
+++ b/aries-backchannels/javascript/server/src/controllers/ConnectionController.ts
@@ -8,9 +8,11 @@ import {
   ConnectionEventTypes,
   ConnectionStateChangedEvent,
   ConnectionState,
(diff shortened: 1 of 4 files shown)


nodlesh (approved):
👍

PR #512, opened 2022-06-15
feat(afj): add v2 credential controller

TimoGlastra:
Add issue credential v2 tests for AFJ and also adds tests for ICv2 interop between AFJ <-> ACA-Py.

Had to extend the ICv2 tests a bit to support using 0160 connections as we don't have didexchange tests yet


nodlesh (approved):
Looks good!

PR #517, opened 2022-06-24
Temporary fix to lock the acapy-resolver to a specific version to enable building the acapy-main test agent

swcurran:


nodlesh (approved):
👍

PR #518, opened 2022-06-24
Pin ACA-Py resolver to v0.1.0 until feature merged into ACA-Py

swcurran:


swcurran:
@dbluhm created a tag for acapy-resolver-universal to make this easier to manage -- pinning to the tag vs. a specific commit.  The features will be merged into ACA-Py Real Soon Now and when that happens, we can drop this requirement in AATH (or adjust as appropriate).

@nodlesh

nodlesh (approved):
👍

Issue #530, opened 2022-07-21
Add script for improved feature/scenario display


tdiesler:
```
$ ./manage tests | grep -A 1 -e @RFC0160

  @T002-RFC0025 @AcceptanceTest @RFC0160
  Scenario Outline: Create 0160 connection between two agents with overlapping transports
```

I'd expect to see all test scenarios displayed that are associated with a given tag. 

The general pattern seems to be ...

```
@RFC0160 @AIP10
Feature: RFC 0160 Aries agent connection functions

   @T001-RFC0160 @critical @AcceptanceTest @MobileTest
   Scenario Outline: establish a connection between two agents
```

which does not list `@T001-RFC0160` for `@RFC0160` nor `@AIP10`

Could this perhaps be fixed by adding the feature tags to their respected nested scenario tags?

tdiesler:
PR #531

nodlesh:
Yes, it would fix the problem adding the feature level tags to the individual test scenarios, however a cleaner approach is to change the manage script to recognize feature level tags and apply them to all scenarios. This way we can take advantage of Behaves tag inheritance in the feature files. Thoughts?

nodlesh:
There also may be a way to accomplish this by leaning on Behave instead of traversing the feature files ourselves. Behave has a [dry-run](https://behave.readthedocs.io/en/stable/behave.html#cmdoption-d) option that may prove handy here.

tdiesler:
I added some comment on [#535](https://github.com/hyperledger/aries-agent-test-harness/pull/535#issuecomment-1192302094)

tdiesler:
@nodlesh are we closing this for now or would you prefer an improved version of `./manage tests` without explicitly duplicating feature tags on their respective scenarios?

tdiesler:
I needed a way to display to display selected features/scenarios without the many of tags i"m not (just yet) interested in - like this ...

```
$ ./manage tests -t @RFC0036     
Selecting by: ['@RFC0036']

@RFC0036
Feature: RFC 0036 Aries agent issue credential
  @T001-RFC0036
  Scenario: Issue a credential with the Holder beginning with a proposal
  @T001.1-RFC0036
  Scenario: Issue a credential with the Holder beginning with a proposal with DID Exchange Connection
  @T002-RFC0036
  Scenario: Issue a credential with the Holder beginning with a proposal with negotiation
  @T003-RFC0036
  Scenario: Issue a credential with the Issuer beginning with an offer
[...]

PR #531, opened 2022-07-21
[#530] Incorrect tests query results for specific AIP or RFC

tdiesler:
This PR generally adds the @RFC and @AIP tags from the "Feature" to the nested "Scenarios"

I assume that this doe not affect test selection - not sure however, if this is true?


nodlesh:
See my comment in #530 
Would you like to take a stab at changing the manage script to recognize feature level tags and apply them to all scenarios in the feature file?

tdiesler:
Yes, that's much improved from last time I looked - we can close this if you like

nodlesh:
Closing as requirement met in #535

Issue #538, opened 2022-07-27
Add general request logging for agent commands


tdiesler:
This would allow us to compare agent command request/response messages for a give test scenario for various agent implementations. It'd be easy to see how the response from one agent impl differs from another

tdiesler:
PR: https://github.com/hyperledger/aries-agent-test-harness/pull/539

This PR produces a `request.log` file like this ...

```
Req: POST http://0.0.0.0:9020/agent/command/connection/create-invitation/ {}
Res: 200 {'connection_id': 'fd1afbb2-cf1f-49ea-861d-ffd89b27e9c1', 'invitation': "{'@id': '75fb5271-2aea-4b39-89f3-5d92b0732cab', '@type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation', 'label': 'aca-py.Acme', 'recipientKeys': ['zFBshFpG82HAUkoWkyqZbkbupCW8sXoDW1KZzMBpXKd'], 'serviceEndpoint': 'http://host.docker.internal:9021'}", 'invitation_url': 'http://host.docker.internal:9021?c_i=eyJAdHlwZSI6ICJkaWQ6c292OkJ6Q2JzTlloTXJqSGlxWkRUVUFTSGc7c3BlYy9jb25uZWN0aW9ucy8xLjAvaW52aXRhdGlvbiIsICJAaWQiOiAiNzVmYjUyNzEtMmFlYS00YjM5LTg5ZjMtNWQ5MmIwNzMyY2FiIiwgImxhYmVsIjogImFjYS1weS5BY21lIiwgInJlY2lwaWVudEtleXMiOiBbInpGQnNoRnBHODJIQVVrb1dreXFaYmtidXBDVzhzWG9EVzFLWnpNQnBYS2QiXSwgInNlcnZpY2VFbmRwb2ludCI6ICJodHRwOi8vaG9zdC5kb2NrZXIuaW50ZXJuYWw6OTAyMSJ9'}
-----
[...]

PR #539, opened 2022-07-27
[#538] Add general request logging for agent commands

tdiesler:


swcurran:
Tried this run `./manage run -d acapy-main -t T002-RFC0036` and didn't get any contents in the `request.log` file. I deleted the .logs folder before the run and confirmed that the folder and `request.log` file was created, but it was not written to during the run.  Is it just that test that I ran, or am I missing something?

Thanks!

tdiesler:
It works for me ...
[request.log](https://github.com/hyperledger/aries-agent-test-harness/files/9281213/request.log)

The logs directory is mapped to the test harness container [like this](https://github.com/tdiesler/aries-agent-test-harness/blob/ghi538/manage#L823) ...

```
-v "$(pwd)/.logs:/aries-test-harness/logs"
```

You could exec into the running harness container and `touch` something in `/aries-test-harness/logs` that should appear in your local .logs directory. I don't think, it has to do with your local python version, because that is fixed in the harness container.

The code that appends to the request.log is [here](https://github.com/tdiesler/aries-agent-test-harness/blob/ghi538/aries-test-harness/agent_backchannel_client.py#L69). You could try ...

```
def request_log(method, agent_url, resp_status, resp_text, payload=''):
    def sorted_payload(val):
        valdict = None
        if isinstance(val, str) and len(val) > 1 and (val[0]+val[-1]) == '{}':
            valdict = dict(json.loads(val))
        elif isinstance(val, dict):
            valdict = val
        if valdict:
            valdict = dict(sorted(valdict.items()))
            val = str({k:sorted_payload(v) for (k,v) in valdict.items()})
        return val
    print(f'Req: {method} {agent_url} {sorted_payload(payload)}')
    print(f'Res: {resp_status} {sorted_payload(resp_text)}')
    print(f'-----')
    if os.path.isdir('/aries-test-harness/logs'):
    ....
```

and see if the log shows on the harness console. In that way, we would know whether the log is called at all and whether this could be a file output issue. Note, the empty request.log is already created by the `.\manage` script - hence it would be there even when `agent_backchannel_client.py` does nothing.

Thanks for looking into this - it'll be interesting to find out why the logs are not showing for you.

tdiesler:
Have you re-built the test harness image?

swcurran:
😊 / facepalm.  Doh...

Looks great.

swcurran (approved):

Issue #543, opened 2022-08-11
Payloads in request log do not preserve json syntax


tdiesler:
Currently, we have json payloads (in request.log) like this ...

```
{
'data': "{
    '@id': '71f6b8b2-02ff-47c2-b241-9d98ac91e410', 
    '@type': 'did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/out-of-band/1.0/invitation', 
    'handshake_protocols': ['did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0'], 
    'label': 'aca-py.Acme', 
    'services': [{
        'id': '#inline', 
        'type': 'did-communication', 
        'recipientKeys': ['did:key:z6MkpjQnfym7n2DcWrVzefJxpMdMz1nTBWCQSv4AGrMpdgcK'], 
        'serviceEndpoint': 'http://host.docker.internal:9021'}], 
    'use_existing_connection': False
    }"
}
```
* invalid single quotes
* json objects as string

tdiesler:
This now changes to (without indents) ...

```
Req: POST http://0.0.0.0:9030/agent/command/out-of-band/receive-invitation/ 
{
  "data" : {
    "@id" : "ebdded7e-b237-46b5-88e4-2deeebe7a825",
    "@type" : "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/out-of-band/1.0/invitation",
    "handshake_protocols" : [ "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/didexchange/1.0" ],
    "label" : "aca-py.Acme",
    "services" : [ {
      "id" : "#inline",
      "type" : "did-communication",
      "recipientKeys" : [ "did:key:z6Mkvu9La9HzqSt9jnivQKNY4znMf7rfcEeTXFuJtk16gbqE" ],
      "serviceEndpoint" : "http://host.docker.internal:9021"
    } ],
    "use_existing_connection" : false
  }
}
```

tdiesler:
This change now also uses pytest to assert the above. You can run it like this ...

```
$ pytest aries-test-harness 
============================================================================================================ test session starts =============================================================================================================
platform darwin -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/tdiesler/git/aries-agent-test-harness, configfile: pytest.ini
collected 1 item                                                                                                                                                                                                                             

[...]

PR #544, opened 2022-08-11
[#543] Payloads in request log do not preserve json syntax

tdiesler:
Signed-off-by: Thomas Diesler <tdiesler@redhat.com>


swcurran (approved):
LGTM -- Thanks!

Issue #456, opened 2022-03-25
issue in aca-py configuration in recent afgo-b-acapy tests


Moopli:
ACA-Py is configured to auto-accept, so it doesn't produce a problem report for a [didexchange exception test](https://allure.vonx.io/api/allure-docker-service/projects/afgo-b-acapy/reports/latest/index.html?redirect=false#behaviors/b42ec1e2628bcda077660120786966a4/865c2ca749632e9b/).

swcurran:
@nodlesh or @ianco -- can you look?  Thanks!

nodlesh:
We are working on removing that auto option in ACA-Py. It shouldn't be on by default.

nodlesh:
The conversation [here](https://github.com/hyperledger/aries-agent-test-harness/pull/448/files#r835644848) is directly related to this.

nodlesh:
I'm assuming all we need to do here is to remove the auto option that remains in the ACA-Py backchannel?

nodlesh:
Closing since #549 has been merged.

PR #548, opened 2022-08-15
Fix for revocation-related tests

ianco:

... also filter out another class of tests not supported by aca-py (tests with no http inbound transport, since the acapy backchannel does't support webhooks over ws)

Note in combination with aca-py PR https://github.com/hyperledger/aries-cloudagent-python/pull/1897 gets us down to one failing test ...


swcurran (approved):

nodlesh:
closes #456

PR #565, opened 2022-09-08
Deal with the new 'deleted' state

ianco:

Not sure if this is optimal ...  ignore the new "deleted" state and/or treat it as a successful credential issue

The alternative (I think) is to turn off "auto delete" for AATH tests so we never delete the cred exchange records (and therefore should never encounter a "deleted" state)


nodlesh (approved):
I think the silent handling of the deleted state in web hooks is acceptable. At least for now.

PR #578, opened 2022-10-20
Fix for when there is no runset that manifests in PASSED or FAILED values

swcurran:


swcurran:
@nodlesh / @ianco -- I'd like to get this merged so that we generate proper results for tonight.

See also the Discord comment about needing to add some new runsets to Allure -- which is why this bug was exposed.

Thanks!

ianco (approved):

Issue #579, opened 2022-10-20
Interop Test Pipeline should check for existing allure projects and create if non-existent


nodlesh:
AS AN Aries Framework Interop Test Developer, I WANT my newly created runset to create the needed project(s) in the allure server, SO THAT there is no waiting period for an admin to create them manually in the allure server.

Currently when a new runset is added that tests interop between aries frameworks it specifies REPORT_PROJECT to push the daily test run results to.  If this doesn't exist it has to manually be created by an admin either by logging into the UI or by making a secure allure api call.
An enhancement could be, inside the script that pushes the results to the Allure server, to call the allure api to check if the project exists on the server before we try to send the results.
`GET https://allure.vonx.io/api/allure-docker-service/projects/search?id=my-project-id`
 If it's not there, use the allure secrets to login as admin and make the api call to create the project.
```
POST GET https://allure.vonx.io/api/allure-docker-service/projects?csrf_access_token=###
{
  "id": "my-project-id"
}
```

The risk here is that someone with a typo in the REPORT_PROJECT might get unneeded or eventually abandoned projects created in the server. This isn't that big of a deal since the admin can go in and delete the project. The smooth operation of new runsets may be of higher benefit than the negative of this risk.

PR #582, opened 2022-10-21
Issue #579 - Creating project before uploading allure results

PatStLouis:
Signed-off-by: Patrick St-Louis <patrick.st-louis@idlab.org>

This PR adds a call to the `/allure-docker-service/projects` api endpoint before uploading the results. If the project already exists, it will simply upload the results to the existing project.

### DIFF
--- a/aries-test-harness/allure/send_results_secure.sh
+++ b/aries-test-harness/allure/send_results_secure.sh
@@ -194,6 +194,13 @@ echo "------------------EXTRACTING-CSRF-ACCESS-TOKEN------------------"
 CRSF_ACCESS_TOKEN_VALUE=$(cat cookiesFile | grep -o 'csrf_access_token.*' | cut -f2)
 echo "csrf_access_token value: $CRSF_ACCESS_TOKEN_VALUE"
 
+echo "------------------CREATE-PROJECT------------------"
+curl -X POST "$ALLURE_SERVER/allure-docker-service/projects" \
+  -H 'Content-Type: application/json' \
+  -H "X-CSRF-TOKEN: $CRSF_ACCESS_TOKEN_VALUE" \
+  -d "{
+    "\""id"\"": "\""$PROJECT_ID"\""
+}" -b cookiesFile -ik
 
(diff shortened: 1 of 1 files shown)


PatStLouis:
@swcurran here is a signed PR. The afgo agent is still having a hard time building, with the following errors.

`# github.com/hyperledger/aries-framework-go/pkg/didcomm/protocol/didexchange`
`../../pkg/didcomm/protocol/didexchange/didex_interop.go:62:17: undefined: didcommutil`
`note: module requires Go 1.19`
`make: *** [Makefile:135: agent-rest-acapy-interop] Error 2`
`The command "/bin/sh -c GO_TAGS=${GO_TAGS} GOPROXY=${GOPROXY} make agent-rest-acapy-interop" returned a non-zero code: 2`
`Docker image build failed.`
`Error: Process completed with exit code 1.`

Would bumping the version of `ARG GO_VER=1.16` to `ARG GO_VER=1.19` in the [Dockerfile.afgo-interop](https://github.com/hyperledger/aries-agent-test-harness/blob/main/aries-backchannels/afgo/Dockerfile.afgo-interop) and [Dockerfile.afgo-master](https://github.com/hyperledger/aries-agent-test-harness/blob/main/aries-backchannels/afgo/Dockerfile.afgo-master) fix this? Not sure if this would break other stuff

swcurran:
Aware of the AF-Go issue -- I've created this issue for it in the AF-Go repo: https://github.com/hyperledger/aries-framework-go/issues/3396

Thanks for the fix!  @nodlesh -- can you please review?

nodlesh (approved):
This is great @PatStLouis. 

So the call to create the project doesn't do anything if it finds it already there? If so, this is much simpler than  I thought, and absolutely meets the criteria for #579.

PatStLouis:
This is what we are using internally and what you describe is the observed behavior. It makes a post request to create the project, if the project already exists it will simply carry on and append the results to the existing project.

In case the project already exists, it will return a 400 `{"message": "project_id '<project_id>' is existent"}`

PR #590, opened 2022-11-03
fix(javascript): remove call to nonce generation method

genaris:
This fixes an import issue in AFJ backchannel that was introduced since aries-framework-javascript [v0.3.0-alpha.8](https://github.com/hyperledger/aries-framework-javascript/releases/tag/v0.3.0-alpha.8).


swcurran (approved):
Tested locally by using runset afj -- worked great.  Thanks!

Issue #622, opened 2022-12-13
Respect bash exit code from a test run for command chaining


tdiesler:
The exit code of a test run is always 0 (i.e. success). It is therefore currently not possible to chain together multiple test runs with the expectation that the latter does not get executed if the former fails - like this ...

```
./manage test -d acapy -b camel -t T005-RFC0036 && ./manage test -d acapy -b camel -t T001-RFC0160
```

tdiesler:
PR: https://github.com/hyperledger/aries-agent-test-harness/pull/623

PR #623, opened 2022-12-13
[#622] Respect bash exit code from a test run for command chaining

tdiesler:
Signed-off-by: Thomas Diesler <tdiesler@redhat.com>


swcurran (approved):
Seems to work -- at least not break things I tried.  A little nervous how this might affect the nightly runs, but we'll see.

PR #630, opened 2022-12-28
fix(javascript): agent type error

genaris:
This small typing workaround fixes the compilation issue that we are seeing since 0.3.0-alpha.40.


nodlesh (approved):
👍

PR #637, opened 2023-01-09
Add the corresponding RFC tag to the revocation feature file

PatStLouis:
I want to use the `-t @RFC0183` tag to select revocation tests.

### DIFF
--- a/.github/workflows/generate-summary-results.yml
+++ b/.github/workflows/generate-summary-results.yml
@@ -10,13 +10,13 @@ jobs:
     if: ((success() || failure()) && ((github.event_name != 'pull_request' && github.repository == 'hyperledger/aries-agent-test-harness')))
     steps:
       - name: checkout-test-harness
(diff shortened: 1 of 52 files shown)


swcurran (approved):

nodlesh:
The RFC tags were/are used to denote where the test was derived from. Very few were derived from RFC0183. Could you not use a combination of tags to achieve the run set?

PatStLouis:
@nodlesh thanks for the clarification. When I want to test a specific feature, it's convenient to tag its RFC and revocation was the only one missing this tag. The distinction between `HIPE0011` and `RFC0183` isn't clear to me yet however the feature file is prepended with 0183 therefore it seemed logical to include this tag for easy access. I know there is the `@revocation` tag however this didn't make sense to me from a programmatical approach (I'm dynamically parsing the features files for visual feedback during a test)
preview

Issue #634, opened 2023-01-06
Run AFJ AATH tests using the Shared Components libraries


TimoGlastra:
Should be quite straightforward once we have support for shared components in AFJ

genaris:
> Should be quite straightforward

Having fun with this right now 😛. Can you assign me?

PR #650, opened 2023-02-22
feat: update Javascript backchannel to use shared components

genaris:
Adjust AFJ backchannel to the new API in 0.4.0.alpha, using AnonCreds module for Present Proof and Issue Credential protocols. 

And most notably: upgrade it to use the new 'shared' components (aries-askar, anoncreds-rs, indy-vdr) as a full replacement of indy-sdk. It still fetches indy-sdk in case we want to run it using it, by setting the environment variable `USE_LEGACY_INDY_SDK`.

Note: in the repo there is a javascript-arm Dockerfile that I suppose it's mostly for development purposes. It has not been yet updated (it will require to use node 18 and probably a few tweaks to make all modules work).

Also note that this PR keeps the test suite as it was before (AIP 1.0 + a couple of protocols). Support for more protocols will be part of another PR (hopefully as soon as we stabilize API for 0.4.0 release).

Solves #634


aries-backchannels/javascript/server/src/TestHarnessConfig.ts
```diff
@@ -55,6 +59,25 @@ export class TestHarnessConfig {
     const agentArgs = await this.getAgentArgs(options)
 
     this._agent = await createAgent(agentArgs)
+
+    try {
+      const key = await this._agent.context.wallet.createKey({keyType: KeyType.Ed25519, privateKey: TypedArrayEncoder.fromString(agentArgs.publicDidSeed)})
```
TimoGlastra, line 64: I was thinking we can now do `agent.wallet.createKey`, but that's in the open PR... nvm

TimoGlastra (approved):
Fantastic!

swcurran:
I see that for the last two days, the AFJ Test Agent has not been starting and so all tests have been failing. I'm assuming that this will fix that issue?

genaris:
> I see that for the last two days, the AFJ Test Agent has not been starting and so all tests have been failing. I'm assuming that this will fix that issue?

Yes! It has been partially broken since a breaking change introduced last week, and completely broken since this Monday, when indy-sdk was removed from AFJ Core. Now we are using shared components by default, so let's see how they behave against other agents!

swcurran:
Not so good with my test locally on the acapy-afj runset (`./manage runset acapy-afj -r` — the -r is for rebuild):

1 feature passed, 6 failed, 7 skipped
7 scenarios passed, 39 failed, 107 skipped
143 steps passed, 39 failed, 1138 skipped, 0 undefined

We’ll see how the nightly run goes.

swcurran:
Much better on the AFJ-AFJ tests. 

5 features passed, 0 failed, 9 skipped
30 scenarios passed, 0 failed, 123 skipped
237 steps passed, 0 failed, 1083 skipped, 0 undefined

PR #655, opened 2023-03-03
Remove the verity backchannel as it is no longer supported

swcurran:
The verify backchannel is no longer being maintained and enquiries to the principals have not received responses.

Removing the backchannel, runsets and references to the backchannel in the generated test run results.


nodlesh (approved):
Looks good.

Issue #658, opened 2023-03-21
AFJ Test Agent Not starting with Askar and Indy VDR config errrors


swcurran:
The AFJ agents are not starting. Example commands

```
./manage build -a javascript
LEDGER_URL_CONFIG=http://test.bcovrin.vonx.io TAILS_SERVER_URL_CONFIG=https://tails.vonx.io ./manage run -d javascript -n -t @RFC0160
```

Output logs:

```
using ngrok end point [bob_agent-ngrok]
Fetching end point from ngrok service
ngrok not ready, sleeping 5 seconds....
Fetching end point from ngrok service
fetched end point [https://64cc-96-54-62-211.ngrok.io]
Starting AFJ agent ...
[2K[1G[1myarn run v1.22.19[22m
[2K[1G[2m$ /src/node_modules/.bin/ts-node src/index.ts -p 9030 -i false[22m
/src/node_modules/ts-node/src/index.ts:744
    return new TSError(diagnosticText, diagnosticCodes);
           ^
TSError: ⨯ Unable to compile TypeScript:
[96msrc/TestAgent.ts[0m:[93m109[0m:[93m18[0m - [91merror[0m[90m TS2554: [0mExpected 1 arguments, but got 0.

[7m109[0m     anoncredsRs: new AnonCredsRsModule(),
[7m   [0m [91m                 ~~~~~~~~~~~~~~~~~~~~~~~[0m

  [96mnode_modules/@aries-framework/anoncreds-rs/build/AnonCredsRsModule.d.ts[0m:[93m6[0m:[93m17[0m
    [7m6[0m     constructor(config: AnonCredsRsModuleConfigOptions);
    [7m [0m [96m                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[0m
    An argument for 'config' was not provided.
[96msrc/TestAgent.ts[0m:[93m110[0m:[93m32[0m - [91merror[0m[90m TS2345: [0mArgument of type '{ networks: [IndyVdrPoolConfig]; }' is not assignable to parameter of type 'IndyVdrModuleConfigOptions'.
  Property 'indyVdr' is missing in type '{ networks: [IndyVdrPoolConfig]; }' but required in type 'IndyVdrModuleConfigOptions'.

[7m110[0m     indyVdr: new IndyVdrModule({
[7m   [0m [91m                               ~[0m
[7m111[0m       networks: [indyNetworkConfig],
[...]

PR #659, opened 2023-03-21
fix(javascript): shared modules constructor parameters

genaris:
Solves #658


swcurran (approved):
LGTM — ran through test locally and looks good.

PR #663, opened 2023-03-30
Reduce frequency of runs to 3 times a week

swcurran:
Given the lack of runners at Hyperledger, suggest we reduce the frequency of runs to 3 times a week from the current daily.  Randomly choos days 1, 3 and 6.


nodlesh (approved):
👍

PR #678, opened 2023-05-06
fix(javascript): node-fetch types

genaris:
Add node-fetch types to allow TypeScript compilation, and also downgrade base image to Ubuntu 20.04, as anoncreds-rs does not require a newer version of GLIBC anymore.


swcurran (approved):
Back to working. Thanks!

PR #680, opened 2023-05-08
Redis cluster for redis persistent queue

Jsyro:
Common component to provide a redis cluster if agents/plugins want to leverage in the future. 

Alternate aca-py image that includes redis-events plugin and plugin config.

To fix: 

- [x] Manage starting/stopping redis-cluster
- [x] Ensuring new network (with subnet) creation works
- [x] ~~plugin startup param in backchannel is conditional~~
- [ ] plugin-config.yml file management, there is a template, but unclear how it is used.


swcurran:
Looks like the `wrapper.sh` script is missing from the `services/redis-cluster` folder, and it is expected.  I tried running `./manage service build redis-cluster` and `./manage service start redis-cluster` and got error `service redis-cluster doesn't exist` for both.

Should add a Markdown document about using the service and invoking the tests.

Adding other comments in the specific files.

aries-backchannels/acapy/acapy_backchannel.py
```diff
@@ -301,6 +301,9 @@ def get_agent_args(self):
         result.append(("--plugin", "universal_resolver"))
         result.append(("--plugin-config", "/data-mount/plugin-config.yml"))
 
+        #HARD CODE REDIS RN
```
swcurran: Does this impact the acapy and acapy-main backchannels?  E.g., is the hard-coded value in this shared script OK?

manage
```diff
@@ -550,7 +550,7 @@ getAgentVersion(){
 
 createNetwork() {
   if [[ -z `docker network ls -q --filter "name=aath_network"` ]]; then
-    docker network create aath_network > /dev/null
+    docker network create aath_network --subnet=172.28.0.0/24 > /dev/null
```
swcurran: @WadeBarnes — is hard-coding this value in this context safe on different docker platforms (especially GHA?). I’m not up on the latest best practices for this.

For background: The goal is that we can have a redis-cluster `service` started up before running tests, and if available one or more of the test-agents can use the redis cluster as they see fit. In our case, we have a variation of the ACA-Py test agent that uses redid.

plugin-config.yml
```diff
@@ -0,0 +1,41 @@
+redis_queue:
```
swcurran, line 1: Is this a duplicate to the `plugin-config.yml` in the `aries-backchannel/acapy` folder?  Is it needed?

aries-backchannels/acapy/acapy_backchannel.py
Jsyro: I think acapy just sees the parameter, but not the plugin, and prints a warning\, but I think I can leave it as is (see the universal_resolver above).

plugin-config.yml
Jsyro: I'm not sure. there is a `plugin-config.template` file in `aries_backchannel/acapy` but i'm not sure how that get's put into the container for runtime.

manage
WadeBarnes: That should work.  It's defining the subnet (address ranges) for the specific named docker network.  It shouldn't conflict with anything in GHA as they typically use the Class A subnet ranges (10.0. 0.0 to 10.255. 255.255).

swcurran:
Something wrong on my system in that I couldn’t get things running — likely something on my end and I’ll try again.

However, did note that `services/redis-cluster/wrpper.sh` is not executable on a Linux system.  I needed to do `chmod ug+x services/redis-cluster/wrpper.sh` to run it.

swcurran:
I'm seeing some weirdness in this PR branch, and I've tried all I can to see what's up. For some reason, on this PR a local ledger is not detected when starting AATH and so the tests don't get started.  A `run` prints `waiting for ledger to start`, waits 60 seconds with "."s printing and then exits.  Things I've tried:

- start local von-network
- run on "main" with `./manage run -d acapy -t ...` and that works fine.
- switch to this PR branch, same command fails as defined
- run same command with `LEDGER_URL_CONFIG=http://test.bcovrin.vonx.io ` prefix on this PR branch and works fine.
- prune docker to make sure it's not some docker weirdness -- same result.

So there seems to be something with the changes in this branch and the detection of the start of the ledger that is not working -- independent of the use of Redis or the ACA-Py Redis instance.

I'll try to look through the changes to see anything...

manage
```diff
@@ -16,18 +16,22 @@ export LEDGER_TIMEOUT=60
 # these two can be overrode via env vars
 export LEDGER_URL_CONFIG="${LEDGER_URL_CONFIG}"
 export TAILS_SERVER_URL_CONFIG="${TAILS_SERVER_URL_CONFIG}"
+export REDIS_CLUSTER_URL_CONFIG="${REDIS_CLUSTER_URL_CONFIG}"
 export AGENT_CONFIG_FILE="${AGENT_CONFIG_FILE}"
 
 # these are derived from the above two
-LEDGER_URL_HOST="${LEDGER_URL_CONFIG:-http://localhost:9000}"
-LEDGER_URL_INTERNAL="${LEDGER_URL_CONFIG:-http://${DOCKERHOST}:9000}"
+LEDGER_URL_HOST="${LEDGER_URL_CONFIG:-http://localhost:9001}"
```
swcurran: OK -- that was easy.  This is the problem.  Why the change to `9001` here?  von-network always runs at 9000.  A quick change and it worked...

manage
```diff
@@ -50,7 +54,7 @@ if [ "${PWD_HOST_FQDN}" != "" ]; then
   fi
   myhost=`ifconfig ${ETH_CONFIG} | grep inet | cut -d':' -f2 | cut -d' ' -f1 | sed 's/\./\-/g'`
   if [ "${GENESIS_URL}" == "" ]; then
-    export GENESIS_URL="http://ip${myhost}-${SESSION_ID}-9000.direct.${PWD_HOST_FQDN}/genesis"
+    export GENESIS_URL="http://ip${myhost}-${SESSION_ID}-9001.direct.${PWD_HOST_FQDN}/genesis"
```
swcurran: Same issue.

manage
```diff
@@ -620,6 +657,16 @@ startServices() {
     fi
   fi
 
+  if [[ -z ${REDIS_CLUSTER_URL_CONFIG} ]]; then
```
swcurran: Redis should probably not be started from here. Either it should be `./manage` command line triggered or it should be started independently and before running `./manage`.  Other services are run that way -- notably `orb`.  I ran a test using the `acapy` backchannel, and redis was started.

manage
Jsyro: Thanks.... my machine has port 9000 occupied, so I've had to keep local changes that change it to 9001, we should try and make it configurable eventually.

manage
Jsyro: I'll look into this, I thought I had replicated the other patterns, but clearly not, thanks.

Jsyro: I agree, it will be unused in most circumstances, it is not started by default anymore. Will need to consider how to notify the user of the situation where the that they are using an image that needs redis but isn't running.

swcurran:
I tried this and had some issues:

* On Linux, the redis-cluster `wrapper.sh` is not executable.  I can address as a follow PR if that is a pain on Windows to do.
* Per #699 — the Dockerfile needs to use `slim-bullseye` vs. just `slim` because of an OS change in `slim:latest`.
* The biggest issue — I couldn’t get the reds-cluster to start.  I ran the command `./manage service start redis-cluster` and the output stopped with all of the nodes “Starting”, as per the image below.  I never did get the service started.
  * Thought it might be a docker thing, so did a `docker system prune` and a `./manage service clean redis-cluster`, but no change when starting the service.

Any ideas?

image

aries-backchannels/acapy/Dockerfile.acapy-redis
```diff
@@ -0,0 +1,36 @@
+FROM python:3.7-slim
```
swcurran: Change to: `FROM python:3.7-slim-bullseye`

services/redis-cluster/README.md
```diff
@@ -0,0 +1,5 @@
+# If you need REDIS
+
+The top level `./manage` script does not stand up the redis cluster by default, it must be started manually. 
+
```
swcurran, line 4: Is this the issue I’m having?  If so, I don’t understand what it is telling me to do. Perhaps be more explicit at telling me what to do vs. pointing to some code?

swcurran:
Progress, but still not working.  I was starting the service up correctly, but the added note in the docs is helpful.  The service is now starting correctly (yay!), and redis starts and stops on command.  No more hanging.

However, the tests are not running properly. I've selected a set of tests that should pass -- the nightly `acapy` runset.

Command: `./manage run -d acapy-redis -t @AcceptanceTest -t ~@wip -t ~@T004-RFC0211 -t ~@DidMethod_orb -t ~@Transport_NoHttpOutbound`

I get timeouts in the tests, such as the following.  Perhaps its a Docker Network thing?  Trying to connect to the wrong IP?

```
@RFC0023 @AIP20
Feature: RFC 0023 Establishing Connections with DID Exchange # features/0023-did-exchange.feature:2
  In order establish a connection with two parties,
  As a responder or a requester,
  I want to use DID Exchange(RFC0023) and Out of Band(RFC0434) protocols to accomplish this.
  @T001-RFC0023 @critical @AcceptanceTest
  Scenario: Establish a connection with DID Exchange between two agents with an explicit invitation  # features/0023-did-exchange.feature:9
    Given we have "2" agents                                                                         # features/steps/0160-connection.py:22 0.000s
      | name | role      |
      | Acme | requester |
      | Bob  | responder |
    When "Bob" sends an explicit invitation to "Acme"                                                # features/steps/0023-did-exchange.py:40 300.793s
      Traceback (most recent call last):
        File "/usr/local/lib/python3.7/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/usr/local/lib/python3.7/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features/steps/0023-did-exchange.py", line 55, in step_impl
          data=data,
        File "/aries-test-harness/agent_backchannel_client.py", line 103, in agent_backchannel_POST
          make_agent_backchannel_request, "POST", agent_url, data=payload
        File "/aries-test-harness/agent_backchannel_client.py", line 41, in run_coroutine_with_kwargs
          return loop.run_until_complete(coroutine(*args, **kwargs))
        File "/usr/local/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
          return future.result()
        File "/aries-test-harness/agent_backchannel_client.py", line 51, in make_agent_backchannel_request
          async with client_session.request(method, path, json=data, params=params) as resp:
        File "/usr/local/lib/python3.7/site-packages/aiohttp/client.py", line 1141, in __aenter__
          self._resp = await self._coro
        File "/usr/local/lib/python3.7/site-packages/aiohttp/client.py", line 560, in _request
          await resp.start(conn)
        File "/usr/local/lib/python3.7/site-packages/aiohttp/client_reqrep.py", line 914, in start
          self._continue = None
        File "/usr/local/lib/python3.7/site-packages/aiohttp/helpers.py", line 721, in __exit__
          raise asyncio.TimeoutError from None
      concurrent.futures._base.TimeoutError

```

Here is from the log of Acme in the `/.logs` folder:

```
Agent running with admin url http://172.17.0.1:9022
ACA-py Backchannel running with ACA-py version: 0.8.2-rc0
Press Ctrl-C to exit ...
2023-06-21 19:37:29,781 aries_cloudagent.core.event_bus ERROR Error occurred while processing event
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/redis/asyncio/connection.py", line 609, in connect
    lambda: self._connect(), lambda error: self.disconnect()
  File "/usr/local/lib/python3.7/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "/usr/local/lib/python3.7/site-packages/redis/asyncio/connection.py", line 648, in _connect
    ssl=self.ssl_context.get() if self.ssl_context else None,
  File "/usr/local/lib/python3.7/asyncio/streams.py", line 77, in open_connection
    lambda: protocol, host, port, **kwds)
  File "/usr/local/lib/python3.7/asyncio/base_events.py", line 962, in create_connection
    raise exceptions[0]
  File "/usr/local/lib/python3.7/asyncio/base_events.py", line 949, in create_connection
    await self.sock_connect(sock, address)
  File "/usr/local/lib/python3.7/asyncio/selector_events.py", line 473, in sock_connect
    return await fut
  File "/usr/local/lib/python3.7/asyncio/futures.py", line 263, in __await__
    yield self  # This tells Task to wait for completion.
  File "/usr/local/lib/python3.7/asyncio/tasks.py", line 318, in __wakeup
    future.result()
  File "/usr/local/lib/python3.7/asyncio/futures.py", line 181, in result
    raise self._exception
  File "/usr/local/lib/python3.7/asyncio/selector_events.py", line 503, in _sock_connect_cb
    raise OSError(err, f'Connect call failed {address}')
TimeoutError: [Errno 110] Connect call failed ('172.96.0.103', 6379)

```

swcurran (approved):
Nice work -- tested.

PR #692, opened 2023-06-05
GENESIS_URL Support for Connectionless proof request on and external ledger

nodlesh:
This PR adds GENESIS_URL support to the test harness. You can now use GENESIS_URL instead of LEDGER_URL_CONFIG when running tests or starting agents as a service. There are currently no test run sets that use the GENESIS_URL. 

### Use Case
The BC Wallet tests in Aries Mobile Test Harness uses AATH agents as a service to accomplish Issuer and Verifier roles with the wallet app tests. Some tests are using an issuer that is on the CANdy network. There was a need to a verifier to do a connectionless proof request with a credential from that issuer. The verifier agent needed to be started with the GENESIS_URL from the CANdy network, and the agent is started in read only mode since it does not need(could not get) write access to the ledger. If read only mode is set, then the ACA-Py agent does not register a did on the network.

There is a new yaml file aries-backchannels/acapy/read_only_ledger_verifier_config.yaml that can be used if anyone wants to start an agent with read only ledger access and with all the auto responding turned on. 

This allows us to do a command as follows to accomplish the use case. 
```
GENESIS_URL=https://whatever.genesis.url.you.want AGENT_CONFIG_FILE=/aries-backchannels/acapy/read_only_ledger_verifier_config.yaml ./manage start -b acapy-main
```

### Other Enhancements
To support the GENESIS_URL above I also refactored how environment variables were gathered and sent to the test containers. There is no longer multiple conditional calls to docker run based on what env vars are set. There is now one call. 

The uniresolver was hardcoded to use the LEDGER_URL. It now will use the GENESIS_URL if it is set. 

### Considerations

- This works for the ACA-Py agent/backchannel. There may need to be work done in other backchannels if they wish to use the GENESIS_URL in their workflows. 

- I tested this with a bunch of different run commands that are used in the interop regression tests. All seemed to be fine since they set the LEDGER_URL_CONFIG in the workflows. However, I did try a run command with no LEDGER_URL or GENESIS_URL. My expectation was that it would start a local ledger, then use that in the tests however there was an issue running a command like this, 
```
./manage run -d acapy-main -t @AcceptanceTest -t @AIP20 -t ~@wip -t @T004-RFC0211
```
The genesis transaction was setup and the agent started but would throw this error. 
```
2023-06-05 18:54:30,258 aries_cloudagent.commands.start ERROR Exception during startup:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/commands/start.py", line 72, in init
    await startup
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/commands/start.py", line 28, in start_app
    await conductor.setup()
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/core/conductor.py", line 185, in setup
    self.root_profile, self.setup_public_did and self.setup_public_did.did
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/config/ledger.py", line 136, in ledger_config
    async with ledger:
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/ledger/indy.py", line 318, in __aenter__
    await self.pool.context_open()
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/ledger/indy.py", line 239, in context_open
    await self.open()
  File "/usr/local/lib/python3.7/site-packages/aries_cloudagent/ledger/indy.py", line 204, in open
    self.handle = await indy.pool.open_pool_ledger(self.name, pool_config)
  File "/usr/local/lib/python3.7/site-packages/indy/pool.py", line 88, in open_pool_ledger
    open_pool_ledger.cb)
  File "/usr/local/lib/python3.7/asyncio/futures.py", line 263, in __await__
    yield self  # This tells Task to wait for completion.
  File "/usr/local/lib/python3.7/asyncio/tasks.py", line 318, in __wakeup
    future.result()
  File "/usr/local/lib/python3.7/asyncio/futures.py", line 176, in result
    raise CancelledError
concurrent.futures._base.CancelledError
2023-06-05 18:54:30,267 aries_cloudagent.indy.sdk.profile DEBUG Profile finalizer called; closing wallet
2023-06-05 18:54:30,324 asyncio ERROR Task exception was never retrieved
future: <Task finished coro=<run_loop.<locals>.done() done, defined at /usr/local/lib/python3.7/site-packages/aries_cloudagent/commands/start.py:77> exception=RuntimeError('cannot reuse already awaited coroutine')>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/asyncio/tasks.py", line 249, in __step
    result = coro.send(None)
RuntimeError: cannot reuse already awaited coroutine
```
I don't believe any of my changes would affect this and I ~~am assuming~~ have confirmed this issue existed before my changes. There are no workflows that use the test harness in this manner as far as I know, so it should not impact the interop regression test cycle.

### DIFF
--- a/manage
+++ b/manage
@@ -13,10 +13,12 @@ SCRIPT_HOME="$( cd "$( dirname "$0" )" && pwd )"
 export COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-aath}"
 export AGENT_TIMEOUT=30
 export LEDGER_TIMEOUT=60
-# these two can be overrode via env vars
+# these can be overridden via env vars
 export LEDGER_URL_CONFIG="${LEDGER_URL_CONFIG}"
 export TAILS_SERVER_URL_CONFIG="${TAILS_SERVER_URL_CONFIG}"
 export AGENT_CONFIG_FILE="${AGENT_CONFIG_FILE}"
+export GENESIS_URL="${GENESIS_URL}"
+export GENESIS_FILE="${GENESIS_FILE}"
 
 # these are derived from the above two
 LEDGER_URL_HOST="${LEDGER_URL_CONFIG:-http://localhost:9000}"
@@ -244,6 +246,29 @@ function initEnv() {
   export RUST_LOG=${RUST_LOG:-warning}
 }
 
+# setup the docker environment variables to be passed to containers
+setDockerEnv() {
+    # set variables that have a different name to what is being set for the container
+  DOCKER_ENV="-e AGENT_NAME=${NAME}"
+  if  [[ -n "$BACKCHANNEL_EXTRA_ARGS" ]]; then
+    DOCKER_ENV="${DOCKER_ENV} -e EXTRA_ARGS=${BACKCHANNEL_EXTRA_ARGS}"
+  fi
+  if  [[ -n "$LEDGER_URL_INTERNAL" ]]; then
+    DOCKER_ENV="${DOCKER_ENV} -e LEDGER_URL=${LEDGER_URL_INTERNAL}"
+  fi
+  if  [[ -n "$TAILS_SERVER_URL_INTERNAL" ]]; then
+    DOCKER_ENV="${DOCKER_ENV} -e TAILS_SERVER_URL=${TAILS_SERVER_URL_INTERNAL}"
+  fi
+
+  # variables that have the same variable name as what is being set for the container
+  declare -a GENERAL_VARIABLES=("DOCKERHOST" "NGROK_NAME" "AIP_CONFIG" "AGENT_CONFIG_FILE" "GENESIS_URL" "GENESIS_FILE")
+  for var in "${GENERAL_VARIABLES[@]}"; do
+    if [[ -n "${!var}" ]]; then
+      DOCKER_ENV+=" -e ${var}=${!var}"
+    fi
+  done
+}
+
 # TODO: set up image builds so you don't need to use `./manage rebuild` to refresh remote source repo
 # - image Dockerfile has an ARG for the commit hash,
 # - build script grabs the HEAD commit hash from the agent's github repo
@@ -304,7 +329,7 @@ pingLedger(){
   ledger_url=${1}
 
   # ping ledger web browser for genesis txns
-  local rtnCd=$(curl -s --write-out '%{http_code}' --output /dev/null ${ledger_url}/genesis)
+  local rtnCd=$(curl -s --write-out '%{http_code}' --output /dev/null ${ledger_url})
   if (( ${rtnCd} == 200 )); then
     return 0
   else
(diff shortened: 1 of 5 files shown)


manage
```diff
@@ -304,7 +329,7 @@ pingLedger(){
   ledger_url=${1}
 
   # ping ledger web browser for genesis txns
-  local rtnCd=$(curl -s --write-out '%{http_code}' --output /dev/null ${ledger_url}/genesis)
+  local rtnCd=$(curl -s --write-out '%{http_code}' --output /dev/null ${ledger_url})
```
swcurran, line 332: Doesn’t this change the definition of “ledger_url”?  The point of ledger URL is that it uses the VON Network ledger browser convention of `/genesis` path returning the ledger file.

swcurran:
Ran into another bug and so was not able to test with this.  A little worried about the line that I commented on. I feel like all of the various ways to pick the location of the genesis file should result in one ENV VAR, and it should be used everywhere.

It’s a little odd that being able to resolve the genesis file is the way to check if the network is running, since for “production” type networks, that’s often on GitHub. On the other hand, if it is a production-like network, we should assume it is always running, so it’s a reasonable assumption.

manage
nodlesh: Line 351 now adds the `/genesis` on on the call to pingUrl if it is a Ledger_URL, otherwise just passes the Genesis URL as is. 
I can change the variable name to just `ledger` instead of `ledger_url`?

nodlesh:
> Ran into another bug and so was not able to test with this. 

What bug did you run into?

nodlesh:
> ... A little worried about the line that I commented on. I feel like all of the various ways to pick the location of the genesis file should result in one ENV VAR, and it should be used everywhere.
>
> It’s a little odd that being able to resolve the genesis file is the way to check if the network is running, since for “production” type networks, that’s often on GitHub. On the other hand, if it is a production-like network, we should assume it is always running, so it’s a reasonable assumption.

I'm not sure I understand how it can be one ENV VAR. I have a Genesis url like this `https://raw.githubusercontent.com/ICCS-ISAC/dtrust-reconu/main/CANdy/test/pool_transactions_genesis` That is much different than a ledger url that we tag `/genesis` on the end of. With one ENV VAR how would I know what to do with it?

I may still be misunderstanding that all these (ledger_url, genesis_url, and genesis_file) are for. Correct me if I'm wrong.  I thought they are all just different ways to do the same thing, get the genesis data for the agent start command. What one you use depends on what are have available to you. For my use case all I have is the genesis_url for CANdy Test. There is no ledger url for me to ping. I have to rely on the genesis_url. Should I be extracting data from the genesis data at the end of the genesis url and pinging that?

swcurran:
Yes — you are right, we need different ways to express the same thing.  What I’m thinking is that externally, any of the three can be used, but internally, they are all converted to a single internal representation. But that may be naive.

The other bug I had was related to von-network — now resolved. Docker images not being rebuilt when needed.

I’ve tested a couple of scenarios (runsets “acapy” and “afgo-acapy”) and all seems fine, but I don’t know if I have triggered the code.  Any other suggestions on what to test, or I’ll just approve/merge.

nodlesh:
If your local tests set a LEDGER_URL_CONFIG then you are hitting the code. 
You can also run the command for my use case to make sure it works off my machine. 
```
GENESIS_URL=https://raw.githubusercontent.com/ICCS-ISAC/dtrust-reconu/main/CANdy/test/pool_transactions_genesis AGENT_CONFIG_FILE=/aries-backchannels/acapy/read_only_ledger_verifier_config.yaml ./manage start -b acapy-main
```

swcurran:
Sorry it took so long for me to get back to this.  I tried the command you suggested and the agent Bob did not start, timing out after 30 seconds. No logs were created, so not quite sure what is happening.

nodlesh:
That is weird. I'm sure you rebuilt acapy-main?

swcurran:
Doh!  :-).  Thanks.  Working. Sorry about that.

swcurran (approved):

nodlesh:
Thanks!

PR #715, opened 2023-07-31
Add --preserve-exchange-records flag to ACA-Py startup

swcurran:

### DIFF
--- a/aries-backchannels/acapy/acapy_backchannel.py
+++ b/aries-backchannels/acapy/acapy_backchannel.py
@@ -237,6 +237,7 @@ class AcaPyAgentBackchannel(AgentBackchannel):
             "--notify-revocation",
             "--open-mediation",
             "--enable-undelivered-queue",
+            "--preserve-exchange-records", # For AATH purposes, exchange records must be retained -- not typical in production
         ]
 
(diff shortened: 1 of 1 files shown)


swcurran:
Added this to fix the breaks in the recent ACA-Py runs caused by the breaking change in ACA-Py to delete exchange records   by default.

@nodlesh -- is there any way to consolidate the places that we have startup settings?  Not sure why we have the settings in multiple ENV files (are they needed?) and we could take the settings out of the Py and put them into one ENV.  Perhaps we could have one ENV with common settings (pulled out of the Python code) and overrides in other settings files for specific tests runs, perhaps associated with specific ACA-Py docker files.  I think (but we should verify) that you can have a setting multiple times with the "last one in" winning.  I'm particularly confused by the "revocation notification" settings, since that seems to for a specific test.

nodlesh:
The environment in the backchannel.py file is the base and the env files are added onto that. 
The extra ENVs were added to support running Agents as a service for AMTH tests. Revocation notification needed to be turned on to test the wallet(s) receiving them.  The ENVs also add all the auto options for the same Mobile test reason. 

I do think it would be cleaner to have a base env file then the custom ones can do as you say (last one in wins).

nodlesh (approved):
LGTM

PR #740, opened 2023-10-26
Fix ACA-Py backchannel to invoke the ACA-Py produced aca-py bin file

swcurran:
Tweaks the current implementation to invoke the ACA-Py produced script instead of bypassing it.  It's not clear to me why the ACA-Py generated script can't be linked from its location and executed through the link, but that doesn't work. Using this approach does seem to work.


swcurran:
Tested by running the equivalent of the GHA -- `./manage runset acapy -r` (`-r` for rebuild).

usingtechnology (approved):
Great solution!

nodlesh (approved):
It works! 🙌

PR #743, opened 2023-11-01
fixed get_acapy_version_as_float function to handle new acapy version

### LINKED ISSUE #742: ACA-Py tests failed on nightly run -- reformat of version?
A recent change in ACA-Py to the version that (it turns out) aligns with the Python PEP for version naming by dropping the "-" when there is a Release Candidate version (e.g. 0.11.0-rc1 becomes 0.11.0rc1).

We'll make that change in the tags, releases and version strings in ACA-Py (pyproject.toml, swagger.json, openapi.json files).

This change has happened, so the fix needs to be made in AATH.

nodlesh:
This PR refactors the get_acapy_version_as_float function to handle the new PEP standard version format without the `-` in the version. Historically it worked with a version number of 0.11.0-rc1 and now it will also work with the version formatted as 0.11.0rc1

closes #742


swcurran (approved):
Verified that the changes work. Main failed, PR works on same `runset`.

PR #749, opened 2023-12-08
fix(javascript): missing supportRevocation parameter

genaris:
Fixing the compilation issue recently thrown as AFJ requires `supportRevocation` parameter to be present when creating credential definition.

Also removing any reference to indy-sdk, whose support in AFJ will be dropped very soon.


swcurran (approved):
Looks great — tested locally.

Thanks!

PR #759, opened 2024-01-08
Add anoncred support

jamshale:
This allows the `askar-anoncreds` wallet and endpoints to be tested in the test harness by running the command `BACKCHANNEL_EXTRA_acapy_main="{\"wallet-type\":\"askar-anoncreds\"}" ./manage run -d acapy-main -t @Anoncreds`

- has an anoncreds context that is set based on the extras env variable.
- It adds a schema and cred def in the anoncred format and passes a anoncred param to the backchannel where the endpoints are changed.
- Adds tags to the tests that are expected to pass. I only added a few tests. `v1` tests are still not passing. I figure if they should then another ticket could build off of this. Along with any other tests we feel are important to work with anoncreds.

Some imports and formatting was changed due to my ruff formatting extension. Not sure if anyone cares. 

Note: This was based off of my acapy fork which includes the recent anoncred endpoint changes. They have been merged to acapy but for some reason acapy main wasn't getting updated with the latest code for me when I was developing.

### DIFF
--- a/aries-backchannels/acapy/acapy_backchannel.py
+++ b/aries-backchannels/acapy/acapy_backchannel.py
@@ -6,40 +6,21 @@ import os
 import signal
 import subprocess
 import sys
-
 from timeit import default_timer
-from typing_extensions import Literal
 from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 
-from aiohttp import (
-    web,
-    ClientSession,
-    ClientRequest,
-    ClientError,
-    ClientTimeout,
-)
-
-from python.agent_backchannel import (
-    AgentBackchannel,
-    AgentPorts,
-    BackchannelCommand,
-    default_genesis_txns,
-    RUN_MODE,
-    START_TIMEOUT,
-)
-from python.utils import flatten, log_msg, output_reader, prompt_loop
-from python.storage import (
-    get_resource,
-    push_resource,
-    pop_resource,
-    pop_resource_latest,
-)
-
 from acapy.routes.agent_routes import routes as agent_routes
-from acapy.routes.mediation_routes import (
-    routes as mediation_routes,
-    get_mediation_record_by_connection_id,
-)
+from acapy.routes.mediation_routes import get_mediation_record_by_connection_id
+from acapy.routes.mediation_routes import routes as mediation_routes
+from aiohttp import (ClientError, ClientRequest, ClientSession, ClientTimeout,
+                     web)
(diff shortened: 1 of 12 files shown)


swcurran (approved):
LGTM.  I had to do a rebuild to get the `-t @anoncreds` to be recognized (probably expected…), but once I did that, the proper tests were executed and succeeded.

nodlesh:
Note that there are other ways to set the wallet-type on an agent besides the EXTRA args. One is to use an ACA-Py config file that contains startup parameters. The other is in a feature file, as seen in the draft PR https://github.com/hyperledger/aries-agent-test-harness/pull/760 where we add start_parameters to the `Given we have n agents` step. I'll probably add support for these other options in that PR.

ianco:
> Note that there are other ways to set the wallet-type on an agent besides the EXTRA args. One is to use an adapt config file that contains startup parameters. The other is in a feature file, as seen in the draft PR #760 where we add start_parameters to the `Given we have n agents` step. I'll probably add support for these other options in that PR.

I think we generally want to set the wallet type globally.  I.e. run me all the tests with aca-py askar wallet, or askar-anoncreds wallet.  I don't see the value in setting this in a feature file where we set this on a per-test basis.

ianco:
Note that the `V1` credential and presentation endpoints are not supported for anoncreds (and won't be) so we should add a tag to these (`@DEPRECATED` or something like that), so we can run `... -t ~@DEPRECATED ...`)

nodlesh:
> I think we generally want to set the wallet type globally. I.e. run me all the tests with aca-py askar wallet, or askar-anoncreds wallet. I don't see the value in setting this in a feature file where we set this on a per-test basis.

Yes, I agree, generally. It depends on if we want these tests to run in existing runsets or have them separate. All I've done here is open up the start parameters to be fully added to the features file data. It only worked with Transport Protocols previously, and instead of just adding the did peer flags, I opened it all up. 

We are adding the `@AnonCreds` tag to tests that need handling for that. I think that should cover the case for the `@Deprecated`, maybe?  

Right now, this PR, checks for wallet type in the EXTRA Args environment variable to set the handling for AnonCreds, I think it also should check if the tag `@AnonCred` is set on the tests.  Whether or not the wallet type setting comes from the feature file (rarely, if ever), or the config file, the tag will handle all cases.

ianco:
`askar-anoncreds` will eventually replace `askar` and we will be deprecating `indy` so we don't need to put too much effort into supporting multiple wallet types.

Agree regarding the `@AnonCreds` tag, however we may deprecate api's for other reasons ...

Issue #696, opened 2023-06-14
Increasing the tests run by AFJ


swcurran:
Currently, the AFJ runset (the tests run through GHActions for AFJ) are limited to those selected as supported back in 2021 or so. Since then Aries Framework JavaScript has been substantially expanded.  I wanted to see what AFJ would look like if (almost) all the AATH tests were run in an all AFJ scenario -- the same set run by ACA-Py.  The following shows the results.

To get many of these to run, some effort is needed:

- Pass through the list of executed tests and see which ones are failing because the AFJ does not support the necessary functionality, and revise the set of tags to be included/excluded.
- Pass through the test results and see which tests are failing because the AFJ backchannel does not support the test, and update the backchannel to enable the test to be run.
- Pass through the failing tests and determine which ones are failing because:
  - The test is wrong
  - The backchannel is wrong
  - AFJ is wrong
  - The RFC upon which the test was written is wrong/unclear

From there, we can update the runsets for AFJ with the other Aries Frameworks and verify/fix the interop between AFJ and other Aries components.

```
Failing scenarios:
  features/0023-did-exchange.feature:9  Establish a connection with DID Exchange between two agents with an explicit invitation
  features/0023-did-exchange.feature:27  Establish a connection with DID Exchange between two agents with an explicit invitation with a public DID
  features/0023-did-exchange.feature:45  Establish a connection with DID Exchange between two agents with an implicit invitation
  features/0023-did-exchange.feature:65  Establish a connection with DID Exchange between two agents with attempt to continue after protocol is completed
  features/0025-didcomm-transports.feature:23  Create DIDExchange connection between two agents with overlapping transports -- @1.1 DIDExchange connection with both agents using HTTP for inbound and outbound transport
  features/0025-didcomm-transports.feature:35  Create DIDExchange connection between two agents with overlapping transports -- @3.1 DIDExchange connection with both agents using WS for inbound and outbound transport, but also supporting http for outbound transport
  features/0025-didcomm-transports.feature:40  Create DIDExchange connection between two agents with overlapping transports -- @4.1 DIDExchange connection with both agents using HTTP and WS for inbound and outbound transport
  features/0025-didcomm-transports.feature:41  Create DIDExchange connection between two agents with overlapping transports -- @4.2 DIDExchange connection with both agents using HTTP and WS for inbound and outbound transport
[...]

PR #760, opened 2024-01-12
Peer DID Interop  tests  - Peer DID Transition

### LINKED ISSUE #752: Add tests for Peer DID Interoperability to AATH to enable a Community Coordinate Update to Qualified DIDs
The Aries Frameworks are trying to converge on the elimination of unqualified peer DIDs, and their replacement with support for various forms of did:peer DIDs.  In this context "peer DIDs" means DIDs used for DIDComm messaging, exchanged by agents when they establish a DIDComm connection (via "Connections" or OOB/DID Exchange protocols). "did:peer" are DIDs defined by according to the [did:peer] specification.

[did:peer]: https://identity.foundation/peer-did-method-spec/index.html

Unqualified DIDs go back to the early days of Aries and the "did:sov" method of defining peer DIDs. An agent, sends to the other agent a "DID" without a `did:XXXX` prefix (hence, unqualified) and a DID Doc. In order to eliminate that, we want to go through a community coordinated update (described in [RFC 0793 unqualified DID transition]) that begins with all Aries frameworks sending out unqualified DIDs, but accepting both unqualified and various qualified `did:peer` DIDs (notably, `did:peer:1`, `did:peer:2` and `did:peer:4`) and later switching to sending out only qualified DIDs.  Currently, all of the Aries Frameworks are being updated up to both accept and send out all types of peer DIDs.

"Accepting" in this case means properly handling a peer DID of a given type sent to it in initiating a connection, and responding by using that DID in completing the connection establishment.  So, if the invitation or request are the first messages in establishing a connection, and they contain a given DID type, the recipient of the invitation or request should respond using the same peer DID type.

[RFC 0793 unqualified DID transition]: https://github.com/hyperledger/aries-rfcs/tree/main/features/0793-unqualfied-dids-transition

The testing we would like is:

- Have the test agents running.
- Have Acme configured to initiate a connection with a specified peer DID type (unqualified, `did:peer:1`, `did:peer:2`, `did:peer:4`)
  - We may need to use the "restart agent with new settings in the middle of a run" approach
- Have Acme connect to Bob, (and perhaps Faber and Mallory)
   - Thought is that by doing three connections, we can have a different test agent (ACA-Py, AFJ, Aries VCX) for each, so fewer test runs, but that may or may not be a good idea.
- Have Acme check that Bob, Faber and Mallory all respond with whatever type of peer DID Acme used.

This task is to create the set of tests, and define how the tests will get run -- e.g. how to on the fly change Acme to initiate connections with the required type of peer DID.

nodlesh:
This PR aims to partially complete #752. This is a draft PR until all ACA-Py tasks are completed below. AFJ tasks and other backchannel updates will be broken out into separate tickets if a ticket doesn't exist for it yet.

- [x] Create Feature File and test definitions for Peer DIDs.
- [x] In this PR, there is also some cleanup of the format of the json output in the ACA-Py Backchannel for readability.
- [x] Add handling to restart an agent with different arguments if the `Given "n" agents` step contains startup args. 
- [x] Get start API call to accept any startup options, including `--emit-peer-did-#`
- [x] In  the ACA-Py Backchannel, make sure args passed in through the start API take precedence over the default args in the backchannel.
- [x] After the feature is finished, change the reset of agents to original configuration to only attempt to restart changed agents instead of all of them. Some of them may not even be running. 
- [x] Add handling of what peer DID method is expected throughout the test and check for those in responses. 
- [x] Add tests that do not set the peer did on the responder and make sure it can respond to the peer DID of the requester.

AFJ Backchannel Updates
- [ ] Implement DID Exchange test harness API routes in the backchannel #696
- [ ] Run DID Exchange tests with AFJ and add to the runsets #696
- [ ] Implement Out of Band test harness routes in the backchannel. #696
- [ ] Run Out of Band tests with AFJ and add to the runsets #696
- [ ] Update the AFJ Backchannel based in the new AnonCreds API #635 
- [ ] Update the restart functions to take any args passed through to start the agent with, including any peer:did setting. 
   - Research is necessary to determine how AFJ sets the emitted peer did. It may not be at the agent startup, therefore we wouldn't need to restart the agent. However other equivalent work would need to be done where ever AFJ is told what peer did method to use.
- [ ] Create runsets for AFJ and ACA-Py to fully exercise qualified Peer DIDs.

VCX Backchannel Updates
- [ ] Apply the modifications above to the VCX Backchannel
   - Research is needed to determine what is exactly required.
- [ ] Create runsets for AFJ, ACA-Py, and VCX to fully exercise qualified Peer DIDs between these agents.

Findy Backchannel Updates?
- [ ] Apply the modifications above to the findy Backchannel?

~~AFGO Backchannel Updates~~
- [ ] ~~Apply the modifications above to the AFGO Backchannel~~


nodlesh:
This PR is ready for review and merge. We may have some help on the AFJ side so it would be good if these enhancements were in main. 

These Qualified DID Tests will run in the regular interop pipeline test execution cycle for ACA-py. It will run in the full acapy runset. It will not run for any other runset until the backchannels have be brought up to scope for these tests. 

Though these tests work with ACA-Py, we expect some of the test specification like the start parameters, etc to change to be more generalized across frameworks once we start working with AFJ and other frameworks in this test context.

dbluhm (dismissed):
LGTM!

aries-test-harness/features/0793-peer-did.feature
```diff
@@ -0,0 +1,96 @@
...
+      And "Acme" receives the response
+      And "Acme" sends complete to "Bob"
+      Then "Acme" and "Bob" have a connection
+
+      Examples:
+         | peer_did_method | start_parameters                                               | responder_start_parameters                                     |
+         | did:peer:1      | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-1"]} | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-3"]} |
+         | did:peer:2      | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-2"]} | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-4"]} |
+         | did:peer:3      | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-3"]} | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-1"]} |
+         | did:peer:4      | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-4"]} | {"wallet-type":"askar-anoncreds", "flags":["emit-did-peer-2"]} |
```
dbluhm: Actually, comments: I'm not sure it makes sense for us to include tests for did:peer:3. Also, ACA-Py has not implemented an `emit-did-peer-1` mode.

nodlesh: I've added them in for when they are implemented these tests should then automatically pass. However, if we prefer, we can just comment them out until or if we support those modes. 
I like a test first approach, where we have failing tests then the software catches up to the tests to make them pass. I can understand that we may not want that here, showing a higher fail rate for unimplemented features.

nodlesh: @swcurran Would you prefer currently unsupported DID Peer Methods to be commented out or leave them failing? I'm inclined to think you would want them commented out.

dbluhm: We are unlikely to implement support for emitting a did:peer:1 or 3

nodlesh: Will any other Aries Framework implement those? The RFC mentions them so that is why I put them in.

dbluhm: AFJ implements 1 but it's actually not possible to exchange just did:peer:3s. You emit 2 and then derive a 3.

nodlesh:
The last commit here comments out the DID Peer Methods that ACA-Py doesn't support. Will add them back if and when we support them or another Aries framework does. 
Also loosened the check for the DID Peer Method to pass of fail a test scenario. The responders DID is just checked to make sure it is qualified, not that it has the same DID Peer Method as the requester.

nodlesh:
@dbluhm or @swcurran Could I get a quick final approval of this? Would like to merge at this point. Thanks.

dbluhm (approved):

swcurran:
Awesome work!  Nice!

PR #793, opened 2024-04-09
fix(javascript): force usage of patched ffi-napi

genaris:
Attempt to quickly fix build issue for AFJ. Couldn't test it running yet due to some issues I have to set-up the network locally.


nodlesh (approved):
This works. Awesome, thanks.

PR #799, opened 2024-04-29
Remove redis plugin install from acapy-main container

nodlesh:
This PR removes the redis queue plugin from being installed in the acapy-main container. It is now as it was before which is only used if someone builds the acapy-redis container. 

This should solve the recent acapy test failures in the AATH Interop test pipeline.

### DIFF
--- a/aries-backchannels/acapy/Dockerfile.acapy-redis
+++ b/aries-backchannels/acapy/Dockerfile.acapy-redis
@@ -17,7 +17,7 @@ COPY python/requirements.txt python/
 COPY acapy/requirements-main.txt acapy/
 
 RUN pip install -r python/requirements.txt -r acapy/requirements-main.txt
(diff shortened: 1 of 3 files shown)


swcurran (approved):
Any idea why it didn’t work?  A bit troubling that the “official” version doesn’t work.

nodlesh:
It does work, if the acapy-redis container is used, the redis plugin is actually installed, and the service is started before running the tests.

nodlesh:
The Acapy backchannel always passes the `--plugin redis...` to the acapy start command. Somewhere along the way, I started installing the plugin in the acapy-main container. When the plugin is not installed, there are no errors when acapy starts even if the `--plugin` is used, it just goes on its way successfully. This was the state of the pipeline for a long time. However, when the plugin is installed, acapy would throw errors if the service was not located, which was the case in the test pipeline over the weekend.

PR #801, opened 2024-05-10
added agent logs in test pipeline

nodlesh:
This PR enables acapy agent logs to be captured in the interop test pipeline. We will be able to see the state of an agent in the container-logs attachment on the run in GitHub. 

There may be a slight risk of disk space on the runner doing this, so we will give it a try and if it is a problem we can revert this change.


swcurran (approved):

swcurran:
Kicked off the tests after merging...

nodlesh:
Awesome, thanks.

PR #843, opened 2024-06-25
Remove unmanaged frameworks

nodlesh:
This PR removes the running of .NET and AFGO frameworks from the interop regression test workflow and the generation of interop results.  The workflows for these frameworks remain but are pre-fixed with `skip-`, which causes them to be ignored by the runner.


gen-interop.sh
```diff
@@ -1,15 +1,13 @@
 #! /bin/bash
 
 # Configuration Data -- order matters in these arrays. A new entry requires an entry in ALL "ta_" arrays
-ta_tlas=("acapy" "afgo" "javascript" "dotnet" "findy" "aries-vcx")
-ta_names=("Aries Cloud Agent Python" "Aries Framework Go" "Aries Framework JavaScript" "Aries Framework .NET" "Findy Agent" "AriesVCX")
-ta_shortnames=("ACA-Py" "AF-Go" "AFJ" "AF-.NET" "Findy" "VCX")
-ta_scopes=("AIP 1, 2" "AIP 2" "AIP 1" "AIP 1" "AIP 1" "AIP 1" "AIP 1")
-ta_exceptions=("None" "None" "Revocation" "Proof Proposal" "Credential Exchange" "Revocation" "Proof Proposals, Public Dids, Revocations")
+ta_tlas=("acapy" "javascript" "findy" "aries-vcx")
```
swcurran: FIndy should be removed asd well. Confirmed that it was archived as of May 28, 2024.

swcurran (changes requested):
Request the removal of Findy as well.

gen-interop.sh
nodlesh: Findy removed.

swcurran (approved):
I’ve added a branch “before-backchannel-deletions” so that we can:

- Remove all of the removed backchannels (and thus, eliminate the dependabot PRs for those backchannels)
- Add a note pointing to the branch so that those backchannels can be seen/revised when needed.

nodlesh:
That's a great idea.

nodlesh:
What should we do with the Allure projects for these frameworks, along with the historical interop summary results? I'm guessing it is safe to blow it all away.

swcurran:
I was assuming we would just leave it, but I’m easy. It’s not hurting anyone, and we won’t be adding to it anymore.

PR #859, opened 2024-08-19
added more permissions to workflow

nodlesh:
This PR is an attempt to fix the `Error: Resource not accessible by integration` issue happening for the last few interop pipeline runs in the generate summary results workflow. If may or may not solve the issue.

### DIFF
--- a/.github/workflows/generate-summary-results.yml
+++ b/.github/workflows/generate-summary-results.yml
@@ -7,6 +7,8 @@ defaults:
 
 permissions:
   contents: write
+  pull-requests: read
+  statuses: write
 
 jobs:
   test:


swcurran (approved):

nodlesh:
Unfortunatley, this PR didn't fix the issue. We are still not getting updates in the interop summary. 
https://github.com/hyperledger/aries-agent-test-harness/actions/runs/10483018155/job/29035027033#step:4:271

swcurran:
Yup.  I think I will work on a separate approach that changes the publishing to use mkdocs.  I’ve been using that for several years, and I think it will be a better approach.  I hope to have that done today or Friday.

nodlesh:
Ok, I was going to add a personal access token to the workflow for the create-pull-request step to see if that works, but I'll hold off if you want to change the approach.

PR #863, opened 2024-08-25
Remove upload logs because action is failing

swcurran:
Removes the action in the Aries VCX jobs to upload the log files because the job is failing.

Not sure why the error is happening, but this action is not part of any other job -- others
are welcome to investigate (I may as well, but I just want the jobs working right now).

Hmmm...seems like just changing the `overwrite` to true might solve it. But are the logs
used at all?

Error:

```
Run actions/upload-artifact@v4
  with:
    name: agent-logs
    path: ./test-harness/.logs/
    if-no-files-found: warn
    compression-level: 6
    overwrite: false
  env:
    DOCKER_CONFIG: /home/runner/work/_temp/docker_login_1724463587634
With the provided path, there will be 5 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

### DIFF
--- a/.github/workflows/test-harness-javascript-ariesvcx.yml
+++ b/.github/workflows/test-harness-javascript-ariesvcx.yml
@@ -37,11 +37,6 @@ jobs:
           TEST_SCOPE: "-t @RFC0036,@RFC0037,@RFC0160 -t ~@T003-RFC0160 -t ~@T004-RFC0160 -t ~@revocation -t ~@RFC0025 -t ~@RFC0183 -t ~@RFC0211 -t ~@RFC0434 -t ~@RFC0453 -t ~@wip -t ~@DIDExchangeConnection -t ~@QualifiedDIDs"
           REPORT_PROJECT: javascript-b-aries-vcx
         continue-on-error: true
-      - uses: actions/upload-artifact@v4
-        with:
-          name: agent-logs
-          path: ./test-harness/.logs/
-      - name: run-send-gen-test-results-secure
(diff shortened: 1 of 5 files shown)


.github/workflows/test-harness-javascript-ariesvcx.yml
```diff
@@ -37,11 +37,6 @@ jobs:
           TEST_SCOPE: "-t @RFC0036,@RFC0037,@RFC0160 -t ~@T003-RFC0160 -t ~@T004-RFC0160 -t ~@revocation -t ~@RFC0025 -t ~@RFC0183 -t ~@RFC0211 -t ~@RFC0434 -t ~@RFC0453 -t ~@wip -t ~@DIDExchangeConnection -t ~@QualifiedDIDs"
           REPORT_PROJECT: javascript-b-aries-vcx
         continue-on-error: true
-      - uses: actions/upload-artifact@v4
-        with:
-          name: agent-logs
-          path: ./test-harness/.logs/
-      - name: run-send-gen-test-results-secure
```
WadeBarnes, line 44: @swcurran, It looks like you deleted one too many lines here.

WadeBarnes (changes requested):

.github/workflows/test-harness-javascript-ariesvcx.yml
swcurran: Doh…good catch. Fixed.

WadeBarnes (approved):

PR #866, opened 2024-09-10
Remove Rev Notification test from credo-acapy runset

nodlesh:
There was a revocation notification test running in the credo-d-acapy runset. It was erroring on the retrieval of the notification by Bob/Holder/Credo. This PR removes that test from the runset. As Credo has no support for the thread_id used by ACA-Py in the revocation notification v1 protocol. 

See
https://credo.js.org/guides/features/aries#divergence-from-aries-rfcs

Quote:

> Revocation Notification v1 uses a different thread_id format ( indy::<revocation_registry_id>::<credential_revocation_id>) than specified in the Aries RFC | Any agents adhering to the revocation notification v1 RFC will not be interoperable with Credo. However, revocation notification is considered an optional portion of revocation, therefore this will not break core revocation behavior. Ideally agents should use and implement revocation notification v2. | Actual implementations (ACA-Py) of revocation notification v1 so far have implemented this different format, so this format change was made to remain interoperable.
> 		
> Revocation Notification v1 uses a different thread_id format ( indy::<revocation_registry_id>::<credential_revocation_id>) than specified in the Aries RFC	Any agents adhering to the [revocation notification v1 RFC](https://github.com/hyperledger/aries-rfcs/tree/main/features/0183-revocation-notification) will not be interoperable with Credo. However, revocation notification is considered an optional portion of revocation, therefore this will not break core revocation behavior. Ideally agents should use and implement revocation notification v2.	Actual implementations (ACA-Py) of revocation notification v1 so far have implemented this different format, so this format change was made to remain interoperable.
>

### DIFF
--- a/.github/workflows/test-harness-acapy-credo.yml
+++ b/.github/workflows/test-harness-acapy-credo.yml
@@ -8,7 +8,7 @@ name: test-harness-acapy-credo
 # This runset uses the current main branch of ACA-Py for all of the agents except Bob (holder),
 # which uses the master branch of Credo TS. The runset covers all of the AIP 1.0 tests
 # except those that are known **not** to work with the Credo TS as the holder,
-# notably those that involve revocation.
+# notably those that involve revocation notification v1.
 #
 # Current
 #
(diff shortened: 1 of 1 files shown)


jamshale (approved):
I'll approve removing the test. I guess a test for v2 RFC 0721 should be created?

Edit: I think I understand better now. The aca-py controller can pick either v1_0 or v2_0 when defining the notification.

swcurran:
Why not change the test to use [V2](https://hyperledger.github.io/aries-rfcs/latest/features/0721-revocation-notification-v2/)?  Or just create another test for v2?

Not part of AIP 2.0, but I think that is OK.

nodlesh:
Yes, absolutely should have a rev notification V2 test. Whether that is this test modified or a new one it really doesn't matter. I'm just trying to understand the existing baseline for credo that was set in the old AFJ runset. It must be the case that this worked at some point in the past with AFJ.

swcurran (approved):

PR #874, opened 2024-09-30
Reverting prompt-toolkit upgrade

swcurran:
Per note from @PatStLouis on the PR that updated the toolkit. Odd that I didn't see an error but I confess I didn't run through all the tests -- just a chunk to see all was well.

### DIFF
--- a/aries-backchannels/python/requirements.txt
+++ b/aries-backchannels/python/requirements.txt
@@ -1,4 +1,4 @@
-prompt_toolkit~=3.0.48
+prompt_toolkit~=2.0.9
 pygments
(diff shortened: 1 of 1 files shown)


PatStLouis:
Here's a test run to highlight this:
https://github.com/hyperledger/aries-cloudagent-python/actions/runs/11114848627/job/30882152164?pr=3261

The error reads:
`ERROR: Cannot install aries-cloudagent and prompt_toolkit~=3.0.48 because these package versions have conflicting dependencies.`

It seems to only affect the BDD Interop Integration Tests.

nodlesh (commented):
👍

nodlesh (approved):

PR #886, opened 2024-10-29
Update manage to stop on agent failure

nodlesh:
When running a set of tests, and the start routine fails to communicate to one of the designated agents, the execution of the tests would continue. Of course this would cause some or all of the tests to fail. This PR changes that logic and when the test harness does not get a successful startup of an agent, will let the user know there was an issue, skip the running of the tests, and stop any running test harness services. This should end the interop results getting complete failures in the reports when it was just an agent startup issue.

### DIFF
--- a/manage
+++ b/manage
@@ -881,12 +881,18 @@ startHarness(){
     else
       waitForAgent Acme ${ACME_ENDPOINT}
     fi
+    if [ $? -eq 1 ]; then
+      return 1
+    fi
   fi
   if [[ "$BOB" != "none" ]]; then
     if [[ IS_BOB_REMOTE -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


swcurran (approved):
Nice fix — that one has bugged me for a while, but I just accepted it :-).

nodlesh:
There is an issue with the credo agent I've been working on for a few days now. So this PR stopped the running of tests last night. However the job to push results to Allure ran and it shows empty in the report. Do you think this is ok, or should we skip the results push if tests don't run. It is an indicator that something went wrong and no tests were executed I suppose. I'm leaning to avoiding a results push if tests are skipped. Agreed?

swcurran:
Yes, I think that makes sense.

PR #890, opened 2024-12-09
OATH Gardening

nodlesh:
This PR is a set of small fixes/enhancements to OATH that I have noticed over the last few weeks. 

- [x] Don't push to Allure reports in pipeline if an agent issue happens and tests don't run.
- [x] Update acapy docker build from aries, and move container to python 3.12.
- [X] Extend wait for agent status to 60 seconds and if agent container shutdown don't let curl hang waiting for a response.


nodlesh:
This is ready. I want to get this into the main repo so I can manually kick off a couple workflows, one passing (acapy) and one failing (credo) to make sure the workflow conditions are correct for submitting to the reporting service. If it works, I'll do another PR to update all workflows with the same logic.

swcurran (approved):

Issue #896, opened 2025-01-14
ACA-Py Test Sets Failing to start...usually


swcurran:
Recently (since Dec. 20) the ACA-Py tests are failing, because the agents are not starting.  Running the tests locally work fine -- agents run and all ACA-Py tests pass.

When I first saw this, I thought it was the unpinned poetry update that broke ACA-Py briefly, but I see that is not the issue.  When I look at the logs from the runs, I see that ngrok gives errors (see attached). However, I checked a successful run from a few days ago (ACA-Py aip10 tests) and the ngrok log looked the same.

[docker-ngrok-tails-server-1.log](https://github.com/user-attachments/files/18403503/docker-ngrok-tails-server-1.log)

swcurran:
@ianco -- could you look into this?  This is pretty high priority -- not having the tests running at all is bad.

ianco:
> @ianco -- could you look into this? This is pretty high priority -- not having the tests running at all is bad.

on it

ianco:
I think the ngrok error is a red herring.  Ngrok changed their model awhile ago to require a "token" - which we don't have - we should probably remove ngrok out of the tails startup ...

swcurran:
Agreed - since it gives the same error on success or failure, it's not like the issue.  Looking forward to determining what it is!

swcurran:
Re-running the test now. Started to yesterday, but stopped because of github issues.  I'll see if anything pops up in that.

ianco:
I'll update the manage script to dump out the aca-py logs if the agent fails to start.  (I'm not having any luck reproducing the error)

ianco:
I restarted a couple of the tests (aca-py aip10 and aip20) and they are both working :-S

I'll check tomorrow to see if any of the "auto" runs failed.

swcurran:
I ran aip20 today, and it did run -- w00t.  However, a number of the tests failed. I'm re-running locally, but pretty sure I ran them yesterday and all passed.  Weird...

ianco:
I'm seeing this error:

`2025-01-15 00:26:11,751 acapy_agent.core.plugin_registry ERROR Module doesn't exist: redis_events.v1_0.redis_queue.events`

ianco:
... not sure why there is a `redis` error when aca-py is starting up (especially since it seems to be intermittent).

We're not actually using (or running) `redis` so maybe we should remove `redis` references from the aca-py backchannel?

@swcurran @jamshale  any thoughts?

[...]

PR #898, opened 2025-01-15
Comment out redis references (not used in OATH testing)

ianco:
Ref issue #896


swcurran (approved):
Presumably to be removed entirely later?

ianco:
> Presumably to be removed entirely later?

Yes, I left it in because (a) I'm not 100% sure this will fix the issue (it's weird that the error is intermittent), and (b) in case someone wants to re-include the redis stuff later (I'm not sure why it was added but then never used; there's a `redis` version of the backchannel, so presumably the redis plug-in should only be included for the redis backchannel?)

PR #900, opened 2025-01-23
Dump agent stdout and stderr directly to file (don't pipe to console)

ianco:
Opening for feedback.

See issue #896 - there is a potential issue with python deadlocking when subprocess output is read using a PIPE.

This PR changes the default behaviour to dump the agent's output directly to a file (actually 2 files) rather than trying to stream to the console.

(It's theoretically possible to stream the file output to the console, although (a) I couldn't get this to work in the context of the backchannel (although I got it working in a stand-alone script), and (b) the file output comes out in big "chunks" anyways (based on the file's buffer size) so output to the console wouldn't be very usable anyways.)

I left the "output streaming" option in place (the default behaviour is the new file output) - it can be selected using a env var `PIPE_AGENT_OUTPUT=True`.

### DIFF
--- a/aries-backchannels/acapy/acapy-main.env
+++ b/aries-backchannels/acapy/acapy-main.env
@@ -1,3 +1,6 @@
 
 LOG_LEVEL=DEBUG
 RUST_LOG=warn
+# set PIPE_AGENT_OUTPUT=True to pipe output from agents "live"
+# (default "False" will dump agent stdout and stderr to separate files)
+# PIPE_AGENT_OUTPUT=True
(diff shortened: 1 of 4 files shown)


ianco:
FYI I still see the "aca-py isn't starting up" error even with this PR (and also with the recent updates to askar or vdr or whichever library had a deadlock issue)

swcurran:
Have you been able to try it with the new Askar (0.4.3)?  Or have you @jamshale ?  I suspect it might be something to do with that.  I think Askar 0.4.3 is in the ACA-Py main...

ianco:
> Have you been able to try it with the new Askar (0.4.3)? Or have you @jamshale ? I suspect it might be something to do with that. I think Askar 0.4.3 is in the ACA-Py main...

Yes this is with the latest Askar

swcurran:
I think the source of the issue (a problem with the universal resolver) has been found and fixed.  I think this was a red herring.

Closing this.  @ianco if you think this should be merged, please reopen.

Issue #872, opened 2024-09-24
AATH support and test expansion for RFC-0771 Anoncred Attachement Formats


nodlesh:
See https://github.com/hyperledger/aries-rfcs/tree/main/features/0771-anoncreds-attachments

ianco:
This will only work with `askar-anoncreds` wallets, no?

Should we update the test runs to use `askar-anoncreds` rather than `askar`?

ianco:
I'm going to put in a check - if the test requires an `askar-anoncreds` wallet and the agent isn't running an `askar-anoncreds` wallet then the test will fail.

For now I'll leave the `anoncreds`-specific tests out of the AcceptanceTests (so they won't get run in the github runs) and we can add them once we set our default wallet type to `askar-anoncreds`.

PR #901, opened 2025-01-28
WIP Implement anoncreds credential format

ianco:
This PR add support for using the `anoncreds` credential format, and adds a test suite for aca-py running `--wallet-type askar-anoncreds`.

See issue #872 

I have updated the tags so that all anoncreds-specific tests are tagged with `@Anoncreds`, and I've added a filter to the existing test runs (using `--wallet-type askar`) to exclude these tests.

I've updated the rest of the tests so that anything using an `indy` credential exchange is tagged `@Indy` (or `@CredFormat_Indy` and these are excluded from the anoncreds test run.

This PR includes a reference to the updated connections plug-in in @jamshale 's branch, so this PR should not be merged until after this connections plug-in is merged (and the relevant files in this PR updated) so that we're not chasing down spurious test errors.

The anoncreds test run looks as follows:

```
LEDGER_URL_CONFIG=http://test.bcovrin.vonx.io TAILS_SERVER_URL_CONFIG=https://tails.vonx.io BACKCHANNEL_EXTRA_acapy_main="{\"wallet-type\":\"askar-anoncreds\"}" ./manage run -d acapy-main -t @AcceptanceTest -t ~@wip -t ~@T004-RFC0211 -t ~@DidMethod_orb -t ~@Transport_NoHttpOutbound -t ~@Indy -t ~@CredFormat_Indy

...

Failing scenarios:
  features/0183-revocation.feature:61  Credential revoked by Issuer and Holder attempts to prove with a prover that doesn't care if it was revoked -- @1.1 

8 features passed, 1 failed, 6 skipped
48 scenarios passed, 1 failed, 118 skipped
372 steps passed, 1 failed, 1073 skipped, 0 undefined
Took 10m35.382s
```

There are still some outstanding issues (can be included in this PR before merging or fixed later):

1. Connections plug-in (as mentioned above).  This plug-in should be completed before this PR is merged (and the reference should be updated to the proper repo).
2. Difference between indy and anoncreds schema and cred def structures. There is some logic in the test harness dealing with the 2 different data formats - this should be cleaned up and a common format used in the test harness (and then only the backchannel needs to deal with the differences).  I ran out of time to fix this ...
3. The above failing test.  May be an aca-py issue, I'll have to leave for @jamshale 
4. Review the scope of tests that an `askar-anoncreds` agent should support.  Right now *all* Indy tests are being excluded.
5. Review scope of other test scenarios (the anoncreds tests are only excluded from the aca-py/aca-py test suites).


ianco:
Question for @swcurran @jamshale - how should we handle tests tagged with `@Anoncreds`?

Some tests (that create schemas, issue credentials, etc) interpret the `@Anoncreds` tag to use the anoncreds credential format, and will only work (with aca-py) with an `askar-anoncreds` wallet.

Some tests (like those in `0793-peer-did.feature`) are tagged with `@Anoncreds` but don't actually have any anoncreds-specific processing (these tests are just establishing connections using various types of peer DIDs).  So these will run regardless of the wallet type.

I'm updating the anoncreds-dependant tests to check that the agent is running with the correct wallet type, and failing with a `400` error `agent wallet cannot support anoncreds format` if not.  (This is done in the backchannel when an anoncreds-specific operation is requested, so some tests tagged with `@Anoncreds` will still run even without an anoncreds wallet.)

But the question is - should we setup a separate set of test suites that use `--wallet-type askar-anoncreds` (vs the existing suites that run `--wallet-type askar`) and then we can filter on the `@Anoncreds` tag as appropriate for each test run?

Or alternately we can force-restart each agent to run `--wallet-type askar-anoncreds` for tests tagged with `@Anoncreds`?  (Personally I don't like this option.)

Right now the `@Anoncreds` tests are passing in the nightly workflows because the `@Anoncreds` tag test isn't being done properly, but this is something I've fixed in this PR :-)

jamshale:
I think I intended for this to be a separate suite to make things easy. But, the tests should in theory be an acapy issuer to credo holder, and either being a verifier in the suite. We don't know if these actually pass at this point, though.

ianco:
Note that some tests are failing when running under an `askar-anoncreds` wallet due to this issue:  https://github.com/openwallet-foundation/acapy/issues/3479

jamshale:
Closing in favor of https://github.com/openwallet-foundation/owl-agent-test-harness/pull/904

PR #902, opened 2025-02-11
Remove universal resolver / Use subprocess.run

jamshale:
This changes to subprocess.run instead of subprocess.Popen allowing log info to be captured on a crash/shutdown. I'm not really sure why Popen was chosen here as the backchannel is only a single process. I haven't seen any side effects.

I removed the universal resolver. There is a universal resolver pinned to a version and started up. However the universal resolver wasn't configured to it and was falling back to the default which is the dev.uniresolver. Hence the issue coming up here.

I believe the uniresolver is only used for the `did:orb` method. I don't know a lot about this and don't think it needs to be tested so I chose to remove it instead of trying to configure it properly in docker. I think it also needs a valid ssl certificate.

Tests will still fail because the connection plugin isn't load yet. I will update https://github.com/openwallet-foundation/owl-agent-test-harness/pull/901 after to try and get everything passing.

### DIFF
--- a/aries-backchannels/acapy/acapy_backchannel.py
+++ b/aries-backchannels/acapy/acapy_backchannel.py
@@ -291,8 +291,6 @@ class AcaPyAgentBackchannel(AgentBackchannel):
             # if the env var is set for tails server then use that.
             result.append(("--emit-new-didcomm-mime-type"))
 
-        result.append(("--universal-resolver"))
-
         # result.append(("--plugin", "redis_events.v1_0.redis_queue.events"))
         result.append(("--plugin-config", "/data-mount/plugin-config.yml"))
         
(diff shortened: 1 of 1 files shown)


esune (approved):
Change looks good/straightforward to me based on @jamshale 's description of teh issue and reasoning

swcurran:
Well…good news and bad.  The good news is that I ran a previously failing RunSet (Credo/ACA-Py) and it now runs — all agents starting.  The bad news is that all of the tests failed…

I did notice that this line is still in the startup up the harness:

```
waiting for uniresolver to start……

```

Should it still be there?

jamshale:
We could remove it as a separate PR. I didn't want to over complicate this one.

I'll check the tests. I ran it from acapy repo and some failed and some passed which I was expecting with the connections protocol removed.

swcurran:
I think the issue with those is that all are using the Connection protocol — take a look at the results of the run: https://allure.vonx.io/allure-docker-service-ui/projects/credo-b-acapy

That would make sense.  So we’ll have to get the connections plugin in use.

jamshale:
Ya. I think the credo <--> acapy all use the v1 protocol currently. I know there was work being done on the credo backchannel.

Suddenly have some issue connecting with the ledger now.

PR #904, opened 2025-02-11
Anoncreds issuance/presentation and connections v1 plugin

jamshale:
This takes over https://github.com/openwallet-foundation/owl-agent-test-harness/pull/901 and just updates the plugin to be installed from the plugin repo.

Note: We'll need to install it for the latest release dockerfile after the release is completed.


swcurran (approved):
Looks great. Running the Credo-ACAPy runset locally. All tests failed without this PR, lots (all?) passing with this PR in place.

This is great news — how easy it is to use the Connections plugin.

Great work!

PR #905, opened 2025-03-13
Create public did with endpoint and remove seeded did creation

jamshale:
This fixes the failed tests for creating a connection with a public did by creating the did:sov with acapy instead of with the seed startup parameter.

The acapy agent will no longer register a did on the ledger before starting the agent and tests. It also no longer uses the insecure seed method at all in the test suite.


jamshale:
See https://allure.vonx.io/allure-docker-service-ui/projects/acapy

swcurran (approved):
Ran locally -- all tests pass.

PR #906, opened 2025-03-13
Repair anoncreds report

jamshale:
This repairs the anoncreds acapy->acapy tests. The env variable wasn't being passed to the run.

Couple other small clean up things.

### DIFF
--- a/.github/workflows/test-harness-acapy-anoncreds.yml
+++ b/.github/workflows/test-harness-acapy-anoncreds.yml
@@ -1,21 +1,4 @@
-name: test-harness-acapy-acapy
-# RUNSET_NAME: "ACA-PY to ACA-Py"
-# Scope: AIP 1.0
-# Exceptions: None
-# SKIP
-# 
-# Summary
-#
-# This runset uses the current main branch of ACA-Py for all of the agents. The runset runs all of the tests in the suite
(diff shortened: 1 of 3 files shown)


.github/workflows/test-harness-acapy-anoncreds.yml
```diff
@@ -1,21 +1,4 @@
-name: test-harness-acapy-acapy
-# RUNSET_NAME: "ACA-PY to ACA-Py"
```
swcurran, line 2: I believe the comments you removed are being pulled in by a script the generates the gh-pages published by this repo.  Suggesting putting them back and updating them to be accurate vs. deleting them.

jamshale: Returned.

swcurran (approved):
Ran this locally in this branch and main and got the same set of tests failing. Is that expected?  E.g. is this fixing the test to run as intended, or to improve the pass rate?

I'm thinking it is to make it run as intended, so approving.

jamshale:
It will fail locally if you run it without the askar-anoncreds wallet type. That's what was happening with the allure reports.

I tested it on my fork and all passed except the couple tests related to the public did that should pass now.

PR #907, opened 2025-03-14
Tests for acapy/credo with anoncreds format

ianco:
This runs a few tests to issue anoncreds format credentials from aca-py to credo.

For acapy/acapy these tests all work (`-d acapy-main`) but with credo they all fail (`-d acapy-main -b credo`).  I'm not smart enough to know why :-S

To run these test with acapy/credo run the following (this combines all the tag filters from the existing credo test runs, with additional tags to run the new tests):

```
BACKCHANNEL_EXTRA_acapy_main="{\"wallet-type\":\"askar-anoncreds\"}" ./manage run -d acapy-main -b credo -t @AcceptanceTest -t @AIP10,@RFC0441,@RFC0211,@T001-RFC0453,@T001-RFC0454b -t ~@wip -t ~@DIDExchangeConnection -t ~@T004-RFC0211 -t ~@QualifiedDIDs -t ~@T001-RFC0183 -t @Anoncreds
```

### DIFF
--- a/.github/workflows/test-harness-acapy-credo-anoncreds.yml
+++ b/.github/workflows/test-harness-acapy-credo-anoncreds.yml
@@ -1,12 +1,12 @@
-name: test-harness-acapy-credo-anoncreds
-# RUNSET_NAME: "ACA-PY to Credo Anoncreds"
-# Scope: Anoncreds
+name: test-harness-acapy-anoncreds-credo
+# RUNSET_NAME: "ACA-PY/AnonCreds to Credo"
+# Scope: AnonCreds
 #
 # Summary
 #
 # This runset uses the current main branch of ACA-Py for all of the agents except Bob (holder),
 # which uses the master branch of Credo TS. The runset covers a small number of credential
-# issue and revocation scenarios for Anoncreds credentials.
+# issue and revocation scenarios for AnonCreds credentials.
 #
 # Current
 #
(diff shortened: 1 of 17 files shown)


swcurran:
Is it worth creating a runset for these, so that they can be run with `./manage runset <name of runset>`.  A runset is a GHA script, where the .manage extracts out the agents and tags to use.  Call it “acapy-credo-anoncreds-rs” maybe?

swcurran:
And, once we get this working, we can add them to the “normal” ACA-Py/Credo tests.

ianco:
> Is it worth creating a runset for these

Yes.  I want to add another test first (with revocation) and I also need to make sure the new (anoncreds) tests are excluded from any existing non-anoncreds test runs.

.github/workflows/test-harness-acapy-credo-anoncreds.yml
```diff
@@ -0,0 +1,55 @@
+name: test-harness-acapy-credo-anoncreds
+# RUNSET_NAME: "ACA-PY to Credo Anoncreds"
```
swcurran: We’re trying to get consistent on using “AnonCreds” vs. “Anoncreds’.  Suggest that update be made.

.github/workflows/test-harness-acapy-credo-anoncreds.yml
```diff
@@ -0,0 +1,55 @@
+name: test-harness-acapy-credo-anoncreds
```
swcurran: How are these different from all the other AnonCreds tests we do?  Is it just that ACA-Py is using the “askar-anoncreds” wallet type?   If so, perhaps rename to “acapy-anoncreds-credo”, since AFAIK there are no special Credo settings for the credx vs. anoncreds-rs libraries.

.github/workflows/test-harness-acapy-credo-anoncreds.yml
swcurran: Per this PR in ACA-Py: https://github.com/openwallet-foundation/acapy/pull/3573

We’ll merge it after the next release.

swcurran (changes requested):
A cleanup request or two and a question.

.github/workflows/test-harness-acapy-credo-anoncreds.yml
ianco: There was no test for AnonCred/Revocation (new test added) and I had to tweak a couple of the other AnonCreds tests to meet some of the tag exclusions in the existing Credo runset.

I'll do the other changes you've requested this afternoon.

jamshale (approved):

swcurran (approved):

PR #909, opened 2025-03-25
Fix issue with un-assigned cred format

ianco:


aries-test-harness/features/steps/0011-0183-revocation.py
```diff
@@ -31,7 +31,11 @@
 def step_impl(context, issuer: str):
 
     issuer_url = context.config.userdata.get(issuer)
-    cred_format = context.rev_cred_format
+    # assume "indy" format unless explicitely specified
```
jamshale: Could use `hasattr` here, but this works fine as well. Could you just fix the spelling for `explicitly` please.

jamshale (approved):

jamshale (approved):

PR #912, opened 2025-03-28
Tutorial for adding VC_DI support

ianco:
Tutorial for adding support for VC_DI - issuing credentials and presenting proofs.  (The section on revocation support is started but is not complete.)  There is an example implementation in my repo (see https://github.com/ianco/owl-agent-test-harness/tree/owl-tutorial-impl) for issuing a credential and presenting a proof (but not for revocation).

### DIFF
--- a/vc_di_tutorial.md
+++ b/vc_di_tutorial.md
@@ -1,0 +1,470 @@
+
+# Tutorial - Adding Data Integrity (VC_DI) Support to OATH
+
+## Background and Relevant Documents
+
+VC_DI support within Aca-Py is detailed here: https://aca-py.org/latest/design/AnoncredsW3CCompatibility/.
+
+For issuing, we use the RFC 0809 VC-DI Attachments (https://identity.foundation/aries-rfcs/latest/features/0809-w3c-data-integrity-credential-attachment/).
+
+For presenting, we use the RFC 0510 DIF Presentation Exchange Attachments (https://identity.foundation/aries-rfcs/latest/aip2/0510-dif-pres-exch-attach/).
+
+
+## Support within Aca-Py Demo Integration Tests
+
+You can run the Aca-Py demo (alice/faber) to see VC_DI credentials in action.
+
+Run the following in two separate shells:
+
+```bash
+AGENT_PORT_OVERRIDE=8010 ./run_demo run faber --wallet-type askar-anoncreds --revocation --cred-type vc_di
+```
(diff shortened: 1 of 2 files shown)


ianco:
> There is an example implementation in my repo

This comparison highlights the code changes required:

https://github.com/ianco/owl-agent-test-harness/compare/owl-tutorial...ianco:owl-agent-test-harness:owl-tutorial-impl?expand=1

vc_di_tutorial.md
```diff
@@ -0,0 +1,613 @@
...
+
+We'll just use existing tests cases, but we'll add support for the VC_DI exchange format.
+
+Let's add VC_DI to a couple of existing AnonCreds-format tests - there is a small subset that have been specifically tested with Aca-Py (and with a credo holder):
+
+```bash
+BACKCHANNEL_EXTRA_acapy_main="{\"wallet-type\":\"askar-anoncreds\"}" ./manage run -d acapy-main -t @AcceptanceTest -t @AIP10,@RFC0441,@RFC0211,@T001-RFC0453,@T001-RFC0454b -t ~@wip -t ~@DIDExchangeConnection -t ~@T004-RFC0211 -t ~@QualifiedDIDs -t ~@T001-RFC0183 -t @Anoncreds
+```
+
+(It's a big set of tags to run only 4 tests, but these are the filters used by the credo runsets.  You can also add `-a credo` to run with credo as the holder (note that these tests currently fail with credo).)
```
swcurran: I think this should be `-b credo` if it is the holder, right?

ianco: good catch!

swcurran:
I believe we want to add these tests to the permenant suite of test cases, don’t we?  If so, how about we do this:

1. Merge this — and perhaps a few updates on the document to make document how to use it (incorporating what I’m proposing).
2. Create a branch at that point, and in the instructions in the document, tell the Tutorial taker to checkout the branch, and then to do the tutorial in their local.  The only thing added to the branch are improvements to the tutorial.
3. After we have that branch, and a fully defined tutorial, we can then add the features that are covered in the tutorial into main, so we have these tests and they are running.

Does that make sense?

Make sense?

ianco:
@swcurran  yes that all makes sense.

(Another improvement required to the document is to make sure the revocation scenario actually works :-P)

swcurran (approved):

PR #915, opened 2025-09-16
Repair manage script arguments

jamshale:
I was looking at the test runs from the acapy repo and noticed that around a month ago they weren't running as long as they should be. The test weren't actually being run because of a `-k` argument not recognized in behave tests. I have no idea why this started happening as it just occurred randomly without any code being merged to either project ?

This `-k` argument shouldn't be here though. the arguments passed in should have the appropriate tags arguments. I think it was added for some legacy interaction with pytest but I'm really not sure.

### DIFF
--- a/manage
+++ b/manage
@@ -1114,12 +1114,12 @@ runTests() {
 
   if [[ "${REPORT}" = "allure" && "${COMMAND}" != "dry-run" ]]; then
       echo "Executing tests with Allure Reports."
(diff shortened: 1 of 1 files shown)


swcurran (approved):

swcurran:
Things seem better -- but not working yet.  I ran two of the test suites, and they seemed to work -- albeit with an exception that seems to be on a specific test. However, no test results were pushed to Allure, and the "Publish Docs" (that picks up the latest results from Allure) did not find any past tests.

Sorry I wasn't monitoring this. I last run the "Publish Docs" on August 12, and all was well. So the problem started sometime after that.  I think Allure keeps 30 days for history, and you can see that only (I think) 2 reports are still there...

jamshale:
Hmm, ok. I didn't think the allure stuff would have been affected by this change. I'll have a closer look.

jamshale:
I think this broke because `behave` library wasn't pinned. There's something going on now with `allure-behave` now which is pinned. It can't parse the tags input properly. I'll figure it out. 

The test do run correctly without allure now.

swcurran:
I think we need to undo this change, to fix the new issue reported -- the test counts being off.  Not sure what is happening, but the Allure results are off, showing, for example, 161 tests were run, when it should be 35.

jamshale:
They are showing up as skipped now instead of actually removing them from the test run. I guess that `-k` option does something in the older version. I'll return it. No need to revert this.

PR #916, opened 2025-09-17
Pin behave to version 1.2.6

jamshale:
I think this should fix the allure problem. Not having behave pinned upgraded it in the docker build to version 1.3.x which apparently has breaking changes related to how the script is parsing the tags.

I tried to allow it to upgrade for a while but the manage script is extremely complicated and I could get the format correct. It's supposed to change to these other tag expressions, but there was also an issue with the argument format.

### DIFF
--- a/aries-test-harness/behave.ini
+++ b/aries-test-harness/behave.ini
@@ -7,6 +7,7 @@
 format=json.pretty
 no_skipped=true
 #logging_level = INFO
+tag_expression_protocol = v1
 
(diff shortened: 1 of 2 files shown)


swcurran (approved):

swcurran:
Looks like that solved the issue. Ran one workflow manually and it pushed results to Allure. None work!

I'll monitor for when the rest of the runs happen -- I think today.

PR #921, opened 2025-12-22
Present Proof V1 Removal

jamshale:
This is for the present proof removal. Making sure all the tests pass. Will need to be updated once the PRs are merge in both acapy and plugins repos.


jamshale:
I think the `--no-deps` requirement is acceptable for 2 plugins both from openwallet-foundation which depend on each other. This shouldn't come up very often. This is a unique circumstance.

swcurran (approved):

PR #925, opened 2026-03-17
Feature/manual bdd test with tags

PatStLouis:
## Summary
Makes the BDD test GitHub Action generic so it can be run manually with any tag set, while keeping the previous default scope for push/PR runs.

## Changes
- **`workflow_dispatch` input:** Adds a `test_tags` input so whoever runs the workflow from the Actions tab can pass a space-separated list of tags (e.g. `@RFC0454 @bbs @wip @aip1`). Supports exclusions with `~@tag`.
- **Tag handling:** A "Build test scope from tags" step converts the input string into the `-t @tag` form expected by the test harness and passes it into the run step.
- **Artifacts:** Agent logs (e.g. `faber_agent.log`, `bob_agent.log`) are still uploaded as a downloadable artifact (`bbs-present-proof-agent-logs`, 14-day retention), including when the run fails.

## How to use
1. Open the **Actions** tab.
2. In the left sidebar, click **Manual BDD test (configurable tags)**.
3. Click **Run workflow** (dropdown on the right).
4. Choose the branch (e.g. `main` or this feature branch).
5. Optionally set **Test tags** (e.g. `@AcceptanceTest ~@wip`).
6. Click the green **Run workflow** button. Download agent logs from **Artifacts** on the run page if needed.

### DIFF
--- a/.github/workflows/test-harness-acapy-bbs-present-proof.yml
+++ b/.github/workflows/test-harness-acapy-bbs-present-proof.yml
@@ -26,8 +26,8 @@ jobs:
   test-bbs-present-proof:
     runs-on: ubuntu-latest
     env:
-      LEDGER_URL_CONFIG: "http://test.bcovrin.vonx.io"
-      GENESIS_URL: "http://test.bcovrin.vonx.io/genesis"
+      LEDGER_URL_CONFIG: "https://test.bcovrin.vonx.io"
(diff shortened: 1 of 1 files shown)


KimEbert42 (approved):
Looks good to me

PatStLouis:
discussed on call today, I will need to make sure latest local updates to this pr are pushed

